PR #19, opened 2020-03-23
Expand Run view paddleboard dropdown, add icons

bodom0015:
## Problem
Multiple upcoming features will require modifying the dropdown menu to include new or expanded items. 
  
## Approach
Preemptively stub-out upcoming methods to (hopefully) prevent a merge conflict.

* Added icons to the Run view dropdown menu
* Reordered dropdown items to match mockups (slight divergence aside)
  * No "Duplicate Tale" in mockups
  * Only one "Export" option going forward
  * Grouped Duplicate/Publish/Export together for now
* Added new Run view dropdown menu items (currently noops/stubs):
  * Save Tale Version
  * Recorded Run
  * Connect to Git repository

## How to Test

4. Navigate to the Run view for the Tale
5. Expand the dropdown at the top-right of the paddleboard
    * You should see a dropdown menu appear
    * Each dropdown item should have an icon
    * You should see (non-functional) options for features that have not yet been implemented
    * The dropdown menu should very nearly match the mockups, seen [here](https://whole-tale.github.io/wholetale-css-mockup/src/run.html)

### DIFF
--- a/src/app/+run-tale/run-tale/run-tale.component.ts
+++ b/src/app/+run-tale/run-tale/run-tale.component.ts
@@ -104,7 +104,7 @@ export class RunTaleComponent extends BaseComponent implements OnInit, OnChanges
 
           return;
         }
-        
+
         this.tale = tale;
         this.logger.info("Fetched tale:", this.tale);
 
@@ -138,6 +138,18 @@ export class RunTaleComponent extends BaseComponent implements OnInit, OnChanges
       this.detectCurrentTab();
     }
 
+    performRecordRun() {
+      console.log('Performing recorded run');
+    }
+
+    saveTaleVersion() {
+      console.log('Saving Tale version');
+    }
+
+    openConnectGitRepoDialog() {
+      console.log('Connecting Git repo');
+    }
+
     rebuildTale(): void {
       const params = { id: this.tale._id };
       this.taleService.taleBuildImage(params).subscribe(res => {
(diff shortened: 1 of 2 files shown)


src/app/+run-tale/run-tale/run-tale.component.ts
```diff
@@ -138,6 +138,18 @@ export class RunTaleComponent extends BaseComponent implements OnInit, OnChanges
       this.detectCurrentTab();
     }
 
+    performRecordRun() {
```
cyberdante, line 141: @bodom0015 - All these functions should be public if you're using them in the template. Otherwise, an `ng build --prod` would fail. Component methods _must be public_ for AOT to work when accessed from the template

bodom0015: Hi @cyberdante,

Thank you for the feedback. From what I understand, TypeScript member functions are [public by default](https://www.typescriptlang.org/docs/handbook/classes.html#public-by-default). I admittedly haven't looked much into AoT or the required patterns there. Are you suggesting that the `public` declaration may be explicitly required in this case to make AoT work properly?

cyberdante: Hey @bodom0015, you're right regarding the default visibility for TypeScript variables and functions, but TypeScript is just the tool for coding our Angular apps. Take a quick look at this [link: AOT compilation](https://blog.mgechev.com/2016/08/14/ahead-of-time-compilation-angular-offline-precompilation/#the-context-property), as it may help. Hit me up should you need an extra pair of eyes on a PR or any other kind of help. I have some time in my hands now that I'm fully working from home.

bodom0015: Ahh, thank you for pointing this out! I see the error now you were talking about:
```bash
ERROR in src/app/+run-tale/run-tale/run-tale.component.html(48,63): : Property 'performRecordedRun' does not exist on type 'RunTaleComponent'. Did you mean 'performRecordRun'?
```

This is indeed a typo that I would have missed, as I usually don't run with `--aot`, and would have caused a runtime error when trying to test #21 

Reading the link you posted, it seems like it's only a problem if we explicitly declare members or methods as `private`. I do have a tendency to put `private readonly` on injected services, but these are typically not bound to directly in templates. I try to take care not to fall into the habit of using `private` outside of that pattern, and will continue to avoid binding directly to injected service state :+1:

PR #54, opened 2020-04-16
Git integration

bodom0015:
## Problem
As a user of the WholeTale platform, it would be nice to be able to integrate my Tale more tightly with Git. This would enable me to more easily track my Tale in SCM tools like GitHub or BitBucket.
  
## Approach
Added inputs for specifying an associated Git repo in two different contexts:
1. during Tale creation 
2. from the Run view's dropdown menu
 
## How to Test  
Prerequisites: running https://github.com/whole-tale/girder_wholetale/pull/430 for API

3. From the Tale Catalog, select Create New... > Create Tale from Git Repository
    * The `create-tale-modal` should display, with an additional input allowing the user to specify a Git repo
4. Enter a Git repo URL/Tale name, choose an environment, and click Create
    * You should be brought to the Run view for your new Tale
5. Navigate to Run > Files > Tale Workspace for this Tale
    * You should see that your git repo has been cloned into the Tale's workspace
6. Navigate to the Tale Catalog and create a new Tale (don't specify a git repo here)
    * You should be brought to the Run view for your new Tale
7. Expand the dropdown at the right side of the Run view and select "Connect to Git repository..."
    * A new modal should display with an input allowing the user to specify a Git repo
    * NOTE: This option still shows up if you have already specified a Git repo.. we have no easy way to detect this from the UI
8. Enter a Git repo URL and submit the form
9. Navigate to Run > Files > Tale Workspace for this Tale
    * You should see that your git repo has been cloned into the Tale's workspace


src/app/+tale-catalog/tale-catalog/tale-catalog.component.ts
```diff
@@ -77,22 +82,46 @@ export class TaleCatalogComponent extends BaseComponent implements AfterViewInit
...
-          this.logger.error("Failed to create Tale:", err);
-        });
+        if (showGit) {
+          // Import Tale from Git repo
+          const params = {
+            url: gitUrl ? gitUrl: '', // Pull from querystring/form
+            imageId: tale.imageId, // Pull from user input
+            asTale: false, // Pull from user input
+            git: gitUrl ? true : false,
+            spawn: true, // ??
```
Xarthisius, line 105: I'd set it to `false` so that behavior matches the regular Tale creation flow.

Xarthisius: Also if we could change view to `/run/<new_tale_id>/` upon submitting modal it would be great. Backend always returns a Tale object immediately regardless of the fact whether importing repo failed or not.

Xarthisius:
Except for error reporting (which potentially is not even a part of this PR) this is good for review for others, right?

src/app/+tale-catalog/tale-catalog/tale-catalog.component.ts
bodom0015: Both excellent points.. updated this PR to set `spawn=false` and to navigate to Tale if one was successfully created.

Also fixed some of the copy-pasta error messages there as well :+1:

bodom0015:
Yep, this should be ready for others to take a look, if they have cycles :+1:

craig-willis:
I've left some related comments on https://github.com/whole-tale/girder_wholetale/pull/430.

Overall the UI looks good. A few comments:
* The API allows me to select a branch via <repo_url>@<branch>, but this isn't clear in the UI.  I think this could easily be addressed with help text under the "Git repository URL" in the UI (e.g., https://github.com/org/repo[@branch])
* As discussed on https://github.com/whole-tale/girder_wholetale/pull/430, I don't think we should enable the Connect menu if the Tale already has a `.git` directory. I realize this is not the conclusion from that discussion.

Xarthisius:
> * As discussed on [whole-tale/girder_wholetale#430](https://github.com/whole-tale/girder_wholetale/pull/430), I don't think we should enable the Connect menu if the Tale already has a `.git` directory. I realize this is not the conclusion from that discussion.

I'd argue this is too costly (would require checking if workspace is a git every time a that menu renders), whereas now you get the error that will say the it's not possible to add a repo.

Xarthisius (approved):

bodom0015:
I agree with @craig-willis... for simplicity's sake, couldn't we keep a string flag called `gitRepo` on the Tale's state for this? We could update the repo URL on the Tale when the clone succeeds (same as we do for publish and elsewhere).

I'm struggling to understand how this solution might complicate things for us, since it shouldn't need to hit the disk/girderfs every time the page loads. Am I missing some bit of complexity here?

ThomasThelen (approved):
I've played around with this a few different ways and things seem to be pretty solid. After the Dec 7 dev call discussion, this gets an OK from me

craig-willis (approved):

PR #55, opened 2020-04-16
Tale versioning panel

bodom0015:
## Problem
As a user of the WholeTale platform, it would be nice to be able to save a particular version/state of my Tale to rerun later. This would enable me to save my favorite / working version of the code to preserve existing evaluation, while continuing to evolve the code in new directions.
  
## Approach
Add a panel to the Run view that allows the user to save a new version of the Tale. This panel will list all existing versions/runs created, and offer functionality to create new versions and runs of different configurations.

## How to Test
Prerequisites: at least one Tale created, newest `deploy-dev` + `wt_versioning` (assumes PRs have been merged)

4. Navigate to the Tale's Run view
    * You should see a new "history" icon on the tab bar at the far right
5. Click on the history icon
    * You should see a timeline appear listing out existing versions for this Tale
    * You should see a button allowing you to "Save Tale Version"
    * [HIDDEN] You should see a dropdown allowing you to choose a Run Configuration
    * [HIDDEN] "Default" should be selected the default selection for Run Configuration  
    * [HIDDEN] You should see a button allowing you to perform a "Recorded Run"
6. [HIDDEN] Expand the "Run Configurations" dropdown
    * You should see a list of existing run configuration
    * You should see a button allowing you to "Edit Run Configurations"
    * TODO: Edit currently does nothing
    * TODO: Wire to API (currently hard-coded / mock data)
7. Click the vertical ellipses beside one of the listed versions
    * You should see a dropdown appear
    * The dropdown should offer options to View Info / Rename / Delete
    * [HIDDEN] The dropdown should offer options to Restore / Compare
    * TODO: Restore / Compare do not yet have API endpoints
8. To test functionality: Save a new Tale version, rename it, view it's info, then delete it


bodom0015:
Updated this to wire up to API:
* list versions on tale-version-panel
* add Create / Rename Version modal
* add Tale Version Info modal
* Delete Version button in dropdown

Also hid unfinished features for now:
* Choose / Edit Run Configurations (API may still be in design)
* Restore / Compare Versions (no API endpoints yet)

ThomasThelen:
I noticed that when I open the versions panel, I get a `Http failure response for https://girder.local.wholetale.org/api/v1/version/root?taleId=5fb045b3dc206f437c4894db: 400 OK` in the notification stream. When looking in the network tab of the debugger, I see the following log in the call to `https://girder.local.wholetale.org/api/v1/version/root?taleId=5fb045b3dc206f437c4894db`

```
{field: "id", message: "Invalid ObjectId: root", type: "validation"}
field: "id"
message: "Invalid ObjectId: root"
type: "validation"
```

I noticed trough the swagger page that there isn't a version/root endpoint listed; I made sure that wt_versioning is up to date. Any ideas?

Xarthisius:
Still testing, but one thing I noticed already: If you don't provide a name for a version in the modal, it fails with 400 cause the request has an empty `name=`. Backend doesn't handle it properly. It expects that argument will not be there at all for autogenerated name.

bodom0015:
@ThomasThelen this was what I pointed out on Monday.

Because there were unmerged PRs in `wt_versioning` (that have now been merged), there were some inconsistencies between the client and server APIs. I'll double-check that everything works as expected now that the merge has been completed.

Xarthisius:
I've been diagnosing issues with girderfs using this PR and noticed weird CSS behavior that I haven't seen before:
dash01
dash02

Xarthisius:
> @Xarthisius ah, I had not noticed this. I'll try to prevent the UI from sending an empty name. Should the backend be updated as well to properly handle this case?

Yeah, there should be a name validation in there, cause it has to be not only non-empty, but also POSIX.

craig-willis:
When I create a version, I'm seeing an error (toast)  "Cannot read property 'createDocumentFragment' of undefined". 

The version appears in the history, but the menu will not display. There are no further errors in the dev console.

bodom0015:
I think I've addressed the CSS issues that @Xarthisius pointed out, but I am not seeing that error message that you're seeing, @craig-willis ...

Good catch tho, as I **do** see the behavior where new versions have dropdowns that don't function until after refresh. If you're still seeing this error, if you have some time we can do some quick debugging to track down where it's happening. I can show you a line you can comment out that will turn that toast into a real stack trace that we can follow.

Xarthisius (approved):

craig-willis:
I'm no longer seeing the above error, so it must've been a local environment issue.  While testing https://github.com/whole-tale/girderfs/pull/18 I noticed that I get a 409 toast if I try to create a version with the same name as an existing version.

bodom0015:
This error toast was by design, although it may need slightly more descriptive text.

While I disagree that version names should be required to be unique, this was previously discussed here: https://github.com/whole-tale/wt_versioning/issues/8

src/app/+run-tale/run-tale/tale-versions-panel/tale-versions-panel.component.html
```diff
@@ -0,0 +1,64 @@
...
+    <div style="margin-top: .25rem" *ngIf="false">
+        <button class="ui button fluid primary" (click)="performRecordedRun()">
+          <i class="fas fa-check-circle icon"></i>
+          Recorded Run
+        </button>
+    </div>
+  </div>
+  <h3>Tale History</h3>
+  <div id="timeline" class="ui middle aligned divided list">
+    <div class="ui placeholder" style="height:100%;" *ngIf="!timeline.length">
```
Xarthisius, line 36: May I suggest: 
```suggestion
    <div class="ui message" style="height:100%;" *ngIf="!timeline.length">
```
in the spirit of https://github.com/whole-tale/ngx-dashboard/pull/114 ?

bodom0015: Committed this externally... I don't trust Github UI and there were other conflicts to fix

craig-willis:
It looks like this PR is now out of sync with https://github.com/whole-tale/girderfs/pull/13 and https://github.com/whole-tale/wt_versioning/pull/12 which drop the `/root` call and change `rootId` to `taleId` for the `/version` and `/version/exists` calls.

Xarthisius:
> It looks like this PR is now out of sync with [whole-tale/girderfs#13](https://github.com/whole-tale/girderfs/pull/13) and [whole-tale/wt_versioning#12](https://github.com/whole-tale/wt_versioning/pull/12) which drop the `/root` call and change `rootId` to `taleId` for the `/version` and `/version/exists` calls.

That's my fault :( Sorry @bodom0015

ThomasThelen:
One thing I noticed is that when entering a name for the version, a request is sent off fairly frequently to `https://girder.local.wholetale.org/api/v1/version/exists?rootId=some_id&name=some_version_name`

While entering in the name 'Test', four requests were sent to that endpoint. Longer names yielded more requests. One way to reproduce it is to open the network tab and slowly type in a version name to see a request for each character.

bodom0015:
No worries :+1: sorry too that I missed this, with the holidays and everything

I've updated this PR to remove the `getRoot` API call, and to change `rootId` to `taleId` throughout

craig-willis (approved):
I've confirmed the latest changes and this looks good.  

I do see @ThomasThelen's issue with the exists check.  I wonder if it would make more sense to do the check on save instead of while typing.  On a slow connection, I'm actually able to hit save before the exist check returns.  That said, I think it should be a separate issue and not block merging this PR.

ThomasThelen (approved):

PR #63, opened 2020-05-05
Set required Tale fields to non-null default values

### LINKED ISSUE #32: Initial tale's author field is not following the schema
**Steps to Reproduce**
1. Deploy and create a new Tale
1. Navigate to the new Tale's *View* > *Metadata*. Click *Edit*
1. Add a description.
1. Hit Save

**Expected Results**
Tale updates.

**Actual Results**
Tale fails to update.

**NOTES**
Looks like initially the Tale is created with `authors: ["Kacper Kowalik"]`, which is not following scheme. As a result PUT is bounced with 400 from the backend.

bodom0015:
## Problem
After creating a Tale, editing/saving metadata does not work properly because some of the fields of the new Tale are invalid by default.

Fixes #32 
  
## Approach
Default all required fields to reasonable defaults.

## How to Test
3. Create a new Tale
    * You should be redirected to the Run > Metadata view for the Tale
4. Scroll down and click the "Edit" button at the bottom
    * An editable form should appear
5. Scroll down and click the "Save" button at the bottom
    * You should receive a toast / popup saying your Tale was saved successfully

### DIFF
--- a/src/app/+run-tale/run-tale/run-tale.component.ts
+++ b/src/app/+run-tale/run-tale/run-tale.component.ts
@@ -31,7 +31,7 @@ enum TaleExportFormat {
 })
 export class RunTaleComponent extends BaseComponent implements OnInit, OnChanges {
     taleId: string;
-    tale: Tale = { title:'???', authors: [], imageId: '', dataSet: [], publishInfo: [] };
+    tale: Tale = { title:'???', authors: [], imageId: '', category: '', license: 'CC-BY-4.0', dataSet: [], description: '', public: false, publishInfo: []  };
     instance: Instance;
     creator: User;
     currentTab = 'metadata';
@@ -104,7 +104,7 @@ export class RunTaleComponent extends BaseComponent implements OnInit, OnChanges
 
           return;
         }
-        
+
         this.tale = tale;
         this.logger.info("Fetched tale:", this.tale);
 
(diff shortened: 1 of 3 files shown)


Xarthisius:
Worth noting that I also hardened the backend [here](https://github.com/whole-tale/girder_wholetale/pull/412) so it's not that easy to create a Tale with invalid values due to the fact that defaults in Girder were wrong.

Xarthisius:
@bodom0015 during `make rebuild_dashboard_next` I get:

```bash
ERROR in src/app/+run-tale/run-tale/run-tale.component.ts(34,5): error TS2739: Type '{ title: string; authors: undefined[]; imageId: string; dataSet: undefined[]; publishInfo: undefined[]; }' is missing the following properties from type 'Tale': category, description, license, public

Makefile:109: recipe for target 'rebuild_dashboard_next' failed
```

cyberdante:
@Xarthisius @bodom0015 - since [_Tale_](https://github.com/whole-tale/ngx-dashboard/blob/cf762f744f474bfb143ab733c5d8693baaa52e93/src/app/tales/models/tale.ts#L6) is an interface, when initializing a variable with this type, you need to provide all fields that are not optional (?). If no value is known at creation time, you must provide a default, but all members must be present anyway.

On a separate note, I saw that you're importing jQuery, which is a library that might be heavy and most likely unnecessary. Angular provides mechanisms to target elements from the DOM (I'm looking at you, [ViewChild](https://stackoverflow.com/a/35209681/3853566)), and for simple element selection by class, vanilla JS could be more suitable than importing a full library:
```
    document.querySelectorAll(".box");
```
is the equivalent of 
```
    // import * as $ from 'jquery';
    declare var $: any;
    // ...
    $('.ui.dropdown').dropdown();
```

bodom0015:
@cyberdante Ah, this is true.. my changes to the `create-tale-modal` included the required values but I had forgotten them in the `run-tale` in the Null Object there.

Adding a commit now to remedy this

PR #66, opened 2020-05-14
Export Tale should download a zipfile

### LINKED ISSUE #49: Exported tale not downloaded
**Steps to Reproduce**
1. Create a Tale
2. Export the Tale

**Expected Results**
Zipfile is downloaded

**Actual Results**
Zipfile is not downloaded (but the content does display in the response)

bodom0015:
## Problem
When a user chooses "Export Tale" from the UI, nothing happens. This is because the `tale/:id/export` does not follow the same patterns as the other endpoints, and requires special handling.
  
Fixes #49 

## Approach
Handle this endpoint by navigating to it directly in the browser. This requires placing our token in the URL as a query parameter.

## How to Test

3. From the Tale Catalog view, hover over a Tale and click "View"
4. At the top-right, expand the dropdown and choose "Export Tale"
    * You may see a new tab open and then immediately close
    * You should see that a download has started for a ZIP file named after the Tale ID

### DIFF
--- a/src/app/+run-tale/run-tale/run-tale.component.ts
+++ b/src/app/+run-tale/run-tale/run-tale.component.ts
@@ -14,6 +14,8 @@ import { enterZone } from '@framework/ngrx/enter-zone.operator';
 import { TaleAuthor } from '@tales/models/tale-author';
 import { routeAnimation } from '~/app/shared';
 
+import { ApiConfiguration } from '@api/api-configuration';
+import { TokenService } from '@api/token.service';
 import { PublishTaleDialogComponent } from './modals/publish-tale-dialog/publish-tale-dialog.component';
 
 // import * as $ from 'jquery';
@@ -46,6 +48,8 @@ export class RunTaleComponent extends BaseComponent implements OnInit, OnChanges
       private taleService: TaleService,
       private instanceService: InstanceService,
       private userService: UserService,
+      private tokenService: TokenService,
+      private config: ApiConfiguration,
       private dialog: MatDialog
     ) {
         super();
@@ -202,9 +206,8 @@ export class RunTaleComponent extends BaseComponent implements OnInit, OnChanges
     }
 
     exportTale(format: TaleExportFormat = TaleExportFormat.BagIt): void {
-      const params = { id: this.tale._id, taleFormat: format };
-      this.taleService.taleExportTale(params).subscribe(res => {
-        this.logger.debug(`Exporting tale=${this.tale._id} to ${format}`, res);
-      });
+      const token = this.tokenService.getToken();
+      const url = `${this.config.rootUrl}/tale/${this.tale._id}/export?token=${token}&taleFormat=${format}`;
+      window.open(url, '_blank');
     }
 }
--- a/src/app/+run-tale/run-tale/run-tale.component.ts
+++ b/src/app/+run-tale/run-tale/run-tale.component.ts
@@ -208,6 +208,6 @@ export class RunTaleComponent extends BaseComponent implements OnInit, OnChanges
     exportTale(format: TaleExportFormat = TaleExportFormat.BagIt): void {
       const token = this.tokenService.getToken();
       const url = `${this.config.rootUrl}/tale/${this.tale._id}/export?token=${token}&taleFormat=${format}`;
-      window.open(url, '_blank');
+      this.windowService.open(url, '_blank');
     }
 }
--- a/src/app/framework/core/models/window.ts
+++ b/src/app/framework/core/models/window.ts
@@ -3,6 +3,6 @@ export interface Window {
   location: any;
 
   alert(msg: string): void;
-
+  open(url: string, target?: string): void;
   confirm(msg: string): void;
 }
--- a/src/app/framework/core/window.service.ts
+++ b/src/app/framework/core/window.service.ts
@@ -4,6 +4,10 @@ export class WindowService implements Window {
   navigator: any = {};
   location: any = {};
 
+  open(url: string, target?: string) {
+    return;
+  }
+
   alert(msg: string): void {
     return;
   }


Xarthisius:
I don't know if it's better or worse, but existing `taleService.taleExportTale` works if you make a blob for download:
```diff
diff --git a/src/app/+run-tale/run-tale/run-tale.component.ts b/src/app/+run-tale/run-tale/run-tale.component.ts
index a4d7d89..c32727c 100644
--- a/src/app/+run-tale/run-tale/run-tale.component.ts
+++ b/src/app/+run-tale/run-tale/run-tale.component.ts
@@ -204,7 +204,11 @@ export class RunTaleComponent extends BaseComponent implements OnInit, OnChanges
     exportTale(format: TaleExportFormat = TaleExportFormat.BagIt): void {
       const params = { id: this.tale._id, taleFormat: format };
       this.taleService.taleExportTale(params).subscribe(res => {
-        this.logger.debug(`Exporting tale=${this.tale._id} to ${format}`, res);
+        const blob = new Blob([res], {
+          type: 'application/zip'
+        });
+        const url = window.URL.createObjectURL(blob);
+        window.open(url);
       });
     }
 }
diff --git a/src/app/api/services/tale.service.ts b/src/app/api/services/tale.service.ts
index d56af8b..0eb0f63 100644
--- a/src/app/api/services/tale.service.ts
+++ b/src/app/api/services/tale.service.ts
@@ -442,7 +442,7 @@ class TaleService extends __BaseService {
     let req = new HttpRequest<any>('GET', this.rootUrl + `/tale/${params.id}/export`, __body, {
       headers: __headers,
       params: __params,
-      responseType: 'json'
+      responseType: 'arraybuffer'
     });
 
     return this.http.request<any>(req).pipe(
```

bodom0015:
@Xarthisius while that would use the existing service code, I am unaware of a way to set the filename using this method without a slight hack of creating a hidden `a` on the page and using JavaScript to click it: https://stackoverflow.com/a/19328891

Without this, the Tale zip downloads as a random UUID (in my case this was `d0527cda-ca98-4a20-8285-34345eef2e35.zip`)

Given this limitation, I'm inclined to leave the current code unless someone knows a cleaner way to set the filename on a Blob download?

cyberdante:
@bodom0015 - you might find this directive useful for making buttons/anchors download certain files on click:
```
import { Directive, Input, HostListener, OnInit } from '@angular/core';
import { StringToBlob, Uint8ArrayToBlob } from '../../classes/utils/file-utils';

@Directive({
    selector: '[downloadable]'
})
export class DownloadFileDirective implements OnInit {
    @Input() public fileName: string;
    @Input() public contentType: string = '';
    @Input() public fileContent: Blob | Uint8Array | string;
    @Input() public preprocessingFunction: (content: Blob | Uint8Array | string) => string;
    @Input() public onComplete: () => void;
    @Input() public onError: (error) => void;
    @Input() public onProcessStarted: () => void;

    constructor() { }

    ngOnInit(): void {

    }

    @HostListener('click', ['$event'])
    onClickEvent(event) {
        if (this.onProcessStarted) {
            this.onProcessStarted();
        }
        try {
            if (typeof this.fileContent === 'string') {
                if (this.preprocessingFunction) {
                    this.fileContent = this.preprocessingFunction(this.fileContent);
                }
                this.fileContent = StringToBlob(this.fileContent as string, this.contentType);
            }
            if (this.fileContent instanceof Uint8Array) {
                this.fileContent = Uint8ArrayToBlob(this.fileContent as Uint8Array, this.contentType);
            }
            if (this.fileContent instanceof Blob && !this.fileContent.size) {
                return;
            }
            if (window.navigator && window.navigator.msSaveOrOpenBlob) {
                // IE11 and Edge
                window.navigator.msSaveOrOpenBlob(this.fileContent, this.fileName);
            } else {
                // Chrome, Safari, Firefox, Opera
                let url = URL.createObjectURL(this.fileContent);
                this.openLink(url);
                // Remove the link when done
                setTimeout(() => window.URL.revokeObjectURL(url), 5000);
            }
        } catch (ex) {
            if (this.onError)
                this.onError(ex);
            throw ex;
        } finally {
            if (this.onComplete)
                this.onComplete();
        }
    }
    private openLink(url: string) {
        let a = document.createElement('a');
        // Firefox requires the link to be in the body
        document.body.appendChild(a);
        a.style.display = 'none';
        a.href = url;
        a.download = this.fileName;
        a.click();
        // Remove the link when done
        document.body.removeChild(a);
    }
}
```
... and these are the functions being referenced in the _import_ section:
```
    export function StringToBlob(byteCharacters: string, contentType = "", sliceSize: number = 1024): Blob {
        const bytesLength = byteCharacters.length;
        const slicesCount = Math.ceil(bytesLength / sliceSize);

        const byteArrays = new Array(slicesCount);
        for (let sliceIndex = 0; sliceIndex < slicesCount; ++sliceIndex) {
            let begin = sliceIndex * sliceSize;
            let end = Math.min(begin + sliceSize, bytesLength);

            let bytes = new Array(end - begin);
            for (let offset = begin, i = 0; offset < end; ++i, ++offset) {
                bytes[i] = byteCharacters[offset].charCodeAt(0);
            }
            byteArrays[sliceIndex] = new Uint8Array(bytes);
        }

        const blob = new Blob(byteArrays, { type: contentType });
        return blob;
    }

    export function Uint8ArrayToBlob(data: Uint8Array, contentType = ""): Blob {
        const blob = new Blob([data], { type: contentType });
        return blob;
    }
```
It can be used like this:
```
<button class="btn btn-primary"
              downloadable
              [fileName]="taleFileName" // <-- this comes from the .ts file
              [contentType]="'application/zip'"
              [fileContent]="taleFileContent"> // <-- this comes from the .ts file
    <i class="fa fa-download"></i> Export Tale
</button>
```

PR #67, opened 2020-05-15
Disable create-tale-modal without name/env

bodom0015:
## Problem
The "Create" button on the `create-tale-modal` is enabled even when the input is invalid. This will likely throw an error that could easily be prevented by disabling the button when necessary.
  
## Approach
Disable the "Create" button if no name has been entered or if no environment has been chosen.

NOTE: This also contains a minor CSS change to add a margin below the search filter

## How to Test
3. Navigate to the Tale Catalog view
4. Click the "Create New Tale" button
    * You should see the Create Tale modal open
    * You should see the "Create" button at the bottom is disabled at first
5. Enter a name and choose an environment
    * You should see that the "Create" button is enabled only after filling both inputs

### DIFF
--- a/src/app/+tale-catalog/tale-catalog/modals/create-tale-modal/create-tale-modal.component.html
+++ b/src/app/+tale-catalog/tale-catalog/modals/create-tale-modal/create-tale-modal.component.html
@@ -34,7 +34,7 @@
 
 <mat-dialog-actions>
   <button class="ui black deny button" mat-dialog-close>Cancel</button>
-  <button class="ui positive right labeled icon button" [disabled]="!newTale.name || !newTale.imageId" [mat-dialog-close]="newTale">
+  <button class="ui positive right labeled icon button" [disabled]="!newTale.title || !newTale.imageId" [mat-dialog-close]="newTale">
     Create New Tale
     <i class="play icon"></i>
   </button>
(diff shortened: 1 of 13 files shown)


src/app/+tale-catalog/tale-catalog/modals/create-tale-modal/create-tale-modal.component.html
```diff
@@ -34,8 +34,8 @@ <h1 mat-dialog-title>Create New Tale</h1>
 
 <mat-dialog-actions>
   <button class="ui black deny button" mat-dialog-close>Cancel</button>
-  <button class="ui positive right labeled icon button" [mat-dialog-close]="newTale">
+  <button class="ui positive right labeled icon button" [disabled]="!newTale.name || !newTale.imageId" [mat-dialog-close]="newTale">
```
Xarthisius, line 37:
```suggestion
  <button class="ui positive right labeled icon button" [disabled]="!newTale.title || !newTale.imageId" [mat-dialog-close]="newTale">
```
otherwise I get 
```text
ERROR in src/app/+tale-catalog/tale-catalog/modals/create-tale-modal/create-tale-modal.component.html(37,57): : Property 'name' does not exist on type 'Tale'.
```

bodom0015:
Thank you, good catch :+1: Sorry for the careless mistake!

PR #69, opened 2020-06-03
View logs modal

### LINKED ISSUE #42: "View Logs" on tasks in notification panel doesn't work
**Steps to Reproduce**
1. Create a Tale
1. Navigate to *View* on that Tale
1. Using vertical ellipsis on the right, select "Rebuild Tale"
1. After a notification shows up (note: I had to refresh the page), click on "View logs"

**Expected Results**
Log modal pops up.

**Actual Results**
Error `t.component.openLogViewerModal is not a function` pops up.

bodom0015:
## Problem
There is no way to view logs of a failed Tale build or job

Fixes #42 
  
## Approach
* Add a log viewer modal that, given a `jobId`, will fetch the logs for that job
* Auto-refresh log data every 2 seconds, cleanup refresh interval OnDestroy

## How to Test
3. Launch any Tale
  * You should see the notification panel automatically display with a new notification
4. Click the "View Logs" button beside the notification
  * You should see a modal display indicating that it is fetching the logs
  * You should see log data appear once it has been fetched
  * You should notice that the log data is automatically refreshes every 2 seconds (this is easier to see this in active builds, but can check Networks tab in your browser's debug console if needed)
5. Click the Close button
  * The modal should close


ThomasThelen:
When running make rebuild_dashboard I stumble across

```
ERROR in src/app/layout/notification-stream/notification-stream.component.html(1,30): : Property 'notificationStream' is private and only accessible within class 'NotificationStreamComponent'.
src/app/layout/notification-stream/notification-stream.component.html(5,44): : Property 'notificationStream' is private and only accessible within class 'NotificationStreamComponent'.
src/app/layout/notification-stream/notification-stream.component.html(5,44): : Property 'notificationStream' is private and only accessible within class 'NotificationStreamComponent'.
src/app/layout/notification-stream/notification-stream.component.html(12,9): : Property 'notificationStream' is private and only accessible within class 'NotificationStreamComponent'.
src/app/layout/notification-stream/notification-stream.component.html(54,29): : Property 'notificationStream' is private and only accessible within class 'NotificationStreamComponent'.
```

bodom0015:
@ThomasThelen good catch! I'm not sure why the compiler has never complained about this before, but I went ahead and made `notificationStream` public :+1:

Xarthisius:
It works as advertised, thanks! I hit a minor issue with a log from a "Create Instance" Job that successfully finished. When you click on "view logs" the spinner in the modal spins indefinitely even though I see the dashboard requesting the data from the server every 2 secs. 

Job object looks like this in this case:
```
log: ["", "", "", ""]
```
I assume the fact it's full of empty strings creates the problem.

bodom0015:
Ah, good catch! I forgot that not all jobs have logs... This highlights a subtle timing issue here where the "latest" job is different based on when you click the "View Logs" button.. The _first_ job (not the last one, as the current code assumes) should be a repo2docker build job, so is it safe to assume that this one should always have logs on it?

Another solution might be to fetch the logs for all `jobId` on the associated notification and just concatenate them all. I can only hope that this wouldn't put much additional load on the server though, but we should only need to auto-refresh the last job in the list (if any).

What do you think?

Xarthisius:
> Another solution might be to fetch the logs for all `jobId` on the associated notification and just concatenate them all. I can only hope that this wouldn't put much additional load on the server though, but we should only need to auto-refresh the last job in the list (if any).

That would require some additional logic to sort them properly (maybe based on `created` timestamp?), but it sounds good to me.

craig-willis:
Building an image the logs work as expected, restarting a tale doesn't.  On restart, the view logs button is enabled but when pressed stays in the "loading" state, likely because there are no logs available from the restart job. The done button in this case doesn't close the modal.

bodom0015:
Good catch on the blank log viewer, @craig-willis! You are definitely correct about the restart job having no logs, but I am unable to reproduce the "Close" button not working in that case. Users should also behind to click anywhere behind the modal to close it as well, seen here: http://recordit.co/Y9TQugGZzv

I did notice that the rebuild notifications are missing the `tale_title` due to a typo, and instead use `title_title`. Fixed the typo here: https://github.com/whole-tale/girder_wholetale/pull/419/files

craig-willis:
I can't explain what happened -- I swear I was seeing both the "loading" and "done" button but now an empty log message and working "close" button as you described. Chalk it up to user error.

PR #73, opened 2020-06-15
Tale popout button

### LINKED ISSUE #44: No indication that tale is stopping
**Steps to Reproduce**
1. Start tale
2. Stop tale

**Expected Results**
Old UI displays "please wait" spinner during stop

**Actual Results**
Stop button remains active, no "please wait" message or other activity indicator. 

**NOTES**
This means that I can press stop multiple times too.

### LINKED ISSUE #46: No "popout" icon on interact
**Steps to Reproduce**
1. Run a tale
2. Select "Interact" if not redirected

**Expected Results**
Popout icon is present on interact panel and selecting it opens the tale in a new tab/window

**Actual Results**
No popout icon is present

**Notes**
I see this isn't on the https://whole-tale.github.io/wholetale-css-mockup/src/run.html, so maybe not a bug but an inconsistency with current UI.

bodom0015:
## Problem
Tale "popout" icon is missing from Run view. Previous dashboard had a button allowing the user to "popout" the iframe into a new tab.

Fixes #46 
Fixes #44 
  
NOTE: The "Stop Tale" button properly shows that the Tale is stopping, but the Run > Interact view does not appear to update correctly

## Approach
* Added the popout button to the tab bar of the Run view
* Improved the status/progress indicators on the Run > Interact view

## How to Test
Prerequisites: at least one Tale created and (ideally) running

3. Choose a running Tale and click "View"
    * You should be brought to the Run > Metadata view
4. Click the "Interact" tab and then click "Run Tale" at the top-right
    * You should see the Tale begin to start
5. Wait for the Tale to finish starting up
    * You should see a blue popout icon on the top-right of the tab bar once the container has started (e.g. when the InstanceStatus is `RUNNING == 1`)
6. Click the popout button
    * You should see a new tab open displaying only the contents within the iframe

### DIFF
--- a/src/app/+run-tale/run-tale/tale-interact/tale-interact.component.html
+++ b/src/app/+run-tale/run-tale/tale-interact/tale-interact.component.html
@@ -6,9 +6,29 @@
     <a class="ui button primary" href="tale.instance.url" target="_blank">Click here to comply</a>
 </div>
 -->
+<pre *ngIf="instance">{{ instance.status }}</pre>
+
 <div class="ui stretched stackable grid" id="tale-interact-container">
   <div class="sixteen wide column" *ngIf="instance">
-    <iframe height="500px" width="100%" [src]="instance.url | safe:'url'" *ngIf="instance && instance.url"></iframe>
+    <!-- Launching state -->
+    <div *ngIf="instance.status === 0" style="margin-top:10vh;text-align:center;">
+      <h3>Tale is launching, please wait...</h3>
+      <i class="fas fa-5x fa-spinner fa-pulse"></i>
+    </div>
+
+    <!-- Running state -->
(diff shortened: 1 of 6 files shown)


src/app/+run-tale/run-tale/tale-interact/tale-interact.component.html
```diff
@@ -6,12 +6,32 @@
     <a class="ui button primary" href="tale.instance.url" target="_blank">Click here to comply</a>
 </div>
 -->
+<pre *ngIf="instance">{{ instance.status }}</pre>
```
Xarthisius, line 9: I just noticed that after I hit the merge button. That's a debug right? If we'd want to show an instance status to the user we would do something more fancy than raw integer above the iframe right?

PR #77, opened 2020-06-26
Fix involatile download links on Run > Metadata view

### LINKED ISSUE #23: Involatile data on metadata page has broken links
**Steps to Reproduce**
1. Deploy and register LIGO
1. Navigate to View > Metadata
1. Click on any file in `Involatile Data` 

**Expected Results**
File should download 

**Actual Results**
Link points to `https://dashboard.local.wholetale.org/null`

bodom0015:
## Problem
Download links under `Involatile Data` are invalid, and contain `null`

Fixes #23 

## Approach
Fix the broken links to use the same API call as the legacy dashboard.

NOTE: Girder's `item/:id/download` link returns a 303 redirect to the file's original location. This can break for particular types of files that the browser can display, which do not appear to be downloaded by default. For example, the `BBH_events_v3.json` from the `LIGO Tutorial` Tale. I believe that sending the `contentDisposition=attached` parameter is supposed to tell Girder to download the item instead of redirecting to it, but I don't think that this parameter is currently being respected in all cases.

## How to Test
Prerequisites: LIGO Tutorial Tale

3. Navigate to the Run > Metadata view for the LIGO tale
4. Scroll down to the *Involatile Data* section
5. Click a few links
    * The `.hd5` files should download as expected
    * NOTE: `BBH_events_v3.json` does not download and simply redirects to the file. This was apparently also true in the legacy dashboard, but was never noticed

### DIFF
--- a/src/app/+run-tale/run-tale/tale-metadata/tale-metadata.component.html
+++ b/src/app/+run-tale/run-tale/tale-metadata/tale-metadata.component.html
@@ -133,7 +133,9 @@
               <b>Involatile Data:</b>
               <span *ngIf="!tale.dataSet.length">No citable data</span>
               <ul *ngIf="tale.dataSet.length">
-                <li *ngFor="let dataset of tale.dataSet; index as i; trackBy: trackById"><a [href]="dataset.doi">{{ dataset.mountPath }}</a></li>
+                <li *ngFor="let dataset of tale.dataSet; index as i; trackBy: trackById">
+                    <a [href]="apiRoot + '/item/' + dataset.itemId + '/download?contentDisposition=attachment'" target="_blank">{{ dataset.mountPath }}</a>
+                </li>
               </ul>
             </div>
             <div class="item">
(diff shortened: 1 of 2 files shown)


Xarthisius:
`apiRoot` is undefined for me. As a result the href I get looks like:

```
https://dashboard.local.wholetale.org/undefined/item/5efa157b258abd042a4555d6/download?contentDisposition=attachment
```

bodom0015:
Whoops, I had somehow missed a file when committing this up. Sorry for the confusion :pray:

I've pushed the missing change, so things should work a bit better now

ThomasThelen:
Looks great! One edge-case is when the external data is a folder. This amounts to the endpoint changing from `item/` to `folder/`, the trick is determining when it's a file or folder.

Steps:
1. Open a Tale
2. Add doi:10.18739/A27H1DM79 to the external data
3. Switch over to the metadata tab
4. Attempt to download it

bodom0015:
Great catch! I had not thought to test folders and the LIGO Tale didn't cover them! :+1:

PR #86, opened 2020-08-14
Fix notification stream reconnect spam

### LINKED ISSUE #5: EventSource reconnect every 45 seconds throws an error
```
[x] Bug report  <!-- Please check the repository for a similar issue or PR before submitting -->
[ ] Feature request
```

**Current behavior**
EventSource API is throwing an error every 45 seconds: "No activity within 45000 milliseconds. Reconnecting." (displayed as an alert)

**Expected/desired behavior**
If a reconnect is an expected event, then perhaps we can detect and squelch this particular message

2. Wait 45 seconds

Current behavior is distracting and annoying

**Environment**
* **Angular version:** 7.2.0-rc0
It is. Trust me.

* **Browser:**
- [x] Chrome (desktop) version XX
- [ ] Chrome (Android) version XX
- [ ] Chrome (iOS) version XX
- [x] Firefox version XX
- [ ] Safari (desktop) version XX
- [ ] Safari (iOS) version XX
- [ ] IE version XX
- [ ] Edge version XX

* **For Tooling issues:**
- Node version: XX  
- Platform: Mac

* Others:
This behavior was present in the EmberJS dashboard as well

bodom0015:
## Problem
Girder's event stream API does not send periodic keep-alive message over the stream. This causes the client to think that the stream is inactive and disconnect every 45000 milliseconds with a thrown error. This error is innocuous, however, since the client immediately reconnects and operations resume as expected.

Note that this issue was also prevalent in the Ember implementation of the dashboard. From the client side, even forcing a disconnect and reconnecting again manually every 30 seconds does not appear to successfully workaround the error message spam. 

Fixes #5 

## Approach
Just watch for the exact error message(s) that are spammed and squelch them. I do not believe that these particular errors would ever be indicative of a real problem, and other notifications appear to still work as intended under these conditions.

Thorough testing may be required to ensure that even more notifications are not being mysteriously skipped or dropped, as has been previously reported. I have been unable to reliably reproduce this behavior.

## How to Test

3. Open your browser's Developer Console to the `Console` tab and wait for a minute or two
    * You should **NOT** see "No activity for 45000 milliseconds. Reconnect" appearing every 45 seconds in the console
4. Navigate to the Tale Catalog, hover over a Tale that you own, and click View
5. At the top-right, click "Run Tale"
    * You should see your Tale begin to start
    * After a moment, you should see the notification stream appear at the top-right
6. Check the developer console again 
    * You still should **NOT** see "No activity for 45000 milliseconds. Reconnect" appearing every 45 seconds in the console


src/app/layout/notification-stream/notification-stream.service.ts
```diff
@@ -11,7 +12,7 @@ import { bypassSanitizationTrustResourceUrl } from '@angular/core/src/sanitizati
 })
 class NotificationStreamService implements OnDestroy {
   static readonly Path = '/notification/stream';
-  static readonly TimeoutMs = 360000;
+  static readonly TimeoutMs = 30;
```
bodom0015, line 15: I will acknowledge that this value for timeout makes absolutely no sense. With a timeout value this low one would expect the client to be constantly reconnecting to the stream, but this does not appear to be the case.

What it actually does (for some extremely odd reason) is that it appears to change the format of the error messages to now provide an `errorCode` in addition to the `errorMessage`. Furthermore, in the reconnect case this `errorMessage` is an empty string, making it trivial to just ignore the empty string and print anything else in order to squelch the error.

PR #94, opened 2020-08-20
Fix: add/style missing info on metadata

### LINKED ISSUE #17: Styling needed for Run > Metadata edit view
```
[ ] Bug report  <!-- Please check the repository for a similar issue or PR before submitting -->
[x] Feature request
```

**Current behavior**
Current styling on Run > Metadata's "edit" style does not match mockups

**Expected/desired behavior**
Replace "edit" mode with a permanently editable form

See Run > Metadata view for a Tale you own, click edit button 

See https://whole-tale.github.io/wholetale-css-mockup/src/run-meta.html

### LINKED ISSUE #47: Missing information on metadata page
**Steps to Reproduce**
Follow "Metadata tab - Owned Tale" test case from test plan

**Expected Results**
Metadata page displays licenses, Created by "your name", published location, can edit tale illustration

**Actual Results**
License dropdown not populated, "Created By" is missing if authors list is empty, no published location, can't edit illustration.

bodom0015:
## Problem
Metadata view is missing some necessary information and styling does not very closely match the mockups.

Fixes #47 
Fixes #17 (but may still need very minor spacing fixes on read-only metadata view)

## Approach
Replace edit/save/cancel with static views that more closely match the mockups provided [here](https://whole-tale.github.io/wholetale-css-mockup/src/run-meta.html).

## How to Test
Prerequisites: at least one Tale created (Owned Tale, does not need to be running), at least one dataset registered

3. Navigate to the Tale Catalog
4. Locate a Tale that you own and click View
5. Click the Files tab to go to the Run > Files view, then attach one or more External Datasets to the Tale
6. Click the Metadata tab to go to the Run > Metadata view
    * You should see a read/write view of the Tale's metadata, allowing you to edit the fields and Save
    * You should see that required `input` fields now show a green border on their left side
    * You should see Citations listed for each of the datasets you've added to the Tale
7. Add an author without filling out Name or ORCID, then scroll to the bottom
    * You should see that invalid fields, such as an `input` that is `required` but empty, are given a red border
    * You should see that the Save button is disabled if a required field is empty
8. Fill in the author's first name, last name, and ORCID
9. Modify the title, category, environment, license, description, and illustration (optional) and click Save
    * You should **not** be able to clear the value of the environment or license dropdowns
    * You should see a notification indicating that the Tale was saved successfully

### DIFF
--- a/src/app/+run-tale/run-tale/tale-metadata/tale-metadata.component.html
+++ b/src/app/+run-tale/run-tale/tale-metadata/tale-metadata.component.html
@@ -3,7 +3,7 @@
     <div class="sixteen wide column">
 
       <!-- Edit Mode (read/write for) -->
-      <div id="taleMetadataForm" class="ui form" *ngIf="editing">
+      <div id="taleMetadataForm" class="ui form" *ngIf="tale._accessLevel > 1">
 
         <div class="inline fields ui grid">
           <label class="two wide column right aligned">Title</label>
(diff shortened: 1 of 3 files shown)


Xarthisius:
I think we're missing the ability to switch between the "edit mode" on the metadata page and a "view mode" to see it rendered as "what would the other people see". I don't see this in the mockups, but I think that without it users will be very confused why metadata pages are so vastly different.

bodom0015:
The "edit mode" was just something I added as a simple way to toggle between the views, but these were never in the mockups. If we do like or prefer this pattern, I can add a "Preview" button (similar to what we have for the `description` textarea) if that might be sufficient?

The "edit mode" was supposed to ease the problem of unsaved edits, but it unfortunately still suffers from the same - a user can make changes, then navigate away during the edit and we will need to discard their changes and/or reload the resource that they were editing.

PR #98, opened 2020-08-31
feat: add move-to-dialog for changing parent folder

### LINKED ISSUE #40: "Move to" on a file yields error
**Steps to Reproduce**
1. Create a Tale
1. Navigate to *View* > *Files* > *Home*.
1. Using `(+)` upload a file.
1. Using action menu on the file, select and click "Move to"

**Expected Results**
Move modal opens

**Actual Results**
`"Cannot read property '_id' of undefined" ` pops up

### LINKED ISSUE #91: Rename "Copy To..." to "Copy" (or enable functionality to select target if it exists)
Bug found during testing of #78

**Steps to Reproduce**
1. Deploy #78 and register Ligo Tale
1. Create a Tale
1. Navigate View Tale > Files > Home
1. Upload a file `test.txt`
1. Using file options (caret) click on "Copy To..."

**Expected Results**
Get a modal allowing to select target where the copy should be created

**Actual Results**
New file `test.txt (1)` is created in a current directory.

**NOTES**
I checked what we do currently in production and we have a "copy" option that behaves the same way. I'm fine with just renaming "Copy To..." to "Copy" if enabling additional modal for destination selection proves to be time consuming.

### LINKED ISSUE #92: Move to... option on files doesn't work
Bug found during testing of #78

**Steps to Reproduce**
1. Deploy #78 and register Ligo Tale
1. Create a Tale
1. Navigate View Tale > Files > Home
1. Upload a file test.txt
1. Using file options (caret) click on "Move To..."

**Expected Results**
Get a modal allowing to select target where the file should be moved

**Actual Results**
Fails with a "Cannot read property '_id' of undefined" message.

bodom0015:
## Problem
The MoveTo dialog was never implemented. There is already some code written that can work as a callback that this modal should execute after submission.

Fixes #40 
Fixes #92 
Fixes #91

## Approach
* Implement the MoveToDialog, allowing user to change the parent folder of a given file or folder

* NOTE: This approach currently encounters issues when moving from a Tale Workspace to Home, and likely vice versa. I may be missing some bits about setting `baseParentId` properly when shifting between data sources. Unless it's an easy fix, this will likely be addressed in a follow-up ticket.

## How to Test
Prerequisites: at least one Tale created, at least two folders created

4. Navigate to Run > Files > Home or Run > Files > Tale Workspace
5. Create two folders here, called "Folder A" and "Folder B"
6. Expand the dropdown beside "Folder B" and choose "Move To..."
    * You should see the MoveToDialog now appear
    * You should be offered a file browser with "Home" and "Tale Workspaces" at the top level
7. Click into Tale Workspaces
    * You should see the names of all Tales listed here
    * Hovering over a Tale name, you should see its ID
8. Double-click (not click) a Tale name to navigate into its Workspace
    * You should **not** be able to select "Tale Workspaces" - this prevents the users from selecting it as a destination
    * You should be brought into the Workspace
    * You should see the Tale's name at the top of the file browser, indicating that you can click it to navigate back up
    * You should not be able to select any files (they are disabled).. only folders are allowed
9. Click the blue header / up arrow twice to go back to the root
10. Double-click Home to navigate into it
    * You should see the contents of your Home folder are listed here
    * You should not be able to select any files (they are disabled).. only folders are allowed
11. Choose / select "Folder A" using single-click, and then click Confirm
    * You should see the modal now close
    * You should see your file disappear from the parent file browser view (it has moved!)
12. Navigate into "Folder A"
    * You should see "Folder B" sitting here

### DIFF
--- a/src/app/+run-tale/run-tale/tale-files/tale-files.component.ts
+++ b/src/app/+run-tale/run-tale/tale-files/tale-files.component.ts
@@ -522,20 +522,22 @@ export class TaleFilesComponent implements OnInit, OnChanges {
   moveElement(event: { element: FileElement; moveTo: FileElement }): void {
     const src = event.element;
     const dest = event.moveTo;
-    const params = { id: src._id, parentId: dest._id };
+    const params = { id: src._id, parentId: dest._id, parentType: ParentType.Folder, baseParentId: dest.baseParentId };
     if (src._modelType === 'folder') {
       // Element is a folder, move it
       this.folderService.folderUpdateFolder(params)
                         .pipe(enterZone(this.zone))
                         .subscribe(resp => {
-        this.logger.debug("Folder moved successfully:", resp);
+        this.logger.info("Folder moved successfully:", resp);
+        this.load();
       });
     } else if (src._modelType === 'item') {
       // Element is an item, move it
       this.itemService.itemUpdateItem(params)
                       .pipe(enterZone(this.zone))
                       .subscribe(resp => {
-        this.logger.debug("Item moved successfully:", resp);
+        this.logger.info("Item moved successfully:", resp);
+        this.load();
       });
     }
   }
--- a/src/app/+run-tale/run-tale/tale-files/tale-files.component.ts
+++ b/src/app/+run-tale/run-tale/tale-files/tale-files.component.ts
@@ -521,9 +521,9 @@ export class TaleFilesComponent implements OnInit, OnChanges {
 
   moveElement(event: { element: FileElement; moveTo: FileElement }): void {
     const src = event.element;
-    const dest = event.moveTo;
-    const params = { id: src._id, parentId: dest._id, parentType: ParentType.Folder, baseParentId: dest.baseParentId };
+    const dest = event.moveTo;;
     if (src._modelType === 'folder') {
+      const params = { id: src._id, parentId: dest._id, parentType: ParentType.Folder, baseParentId: dest.baseParentId }
       // Element is a folder, move it
       this.folderService.folderUpdateFolder(params)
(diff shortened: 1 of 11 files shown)


src/app/+run-tale/run-tale/tale-files/tale-files.component.ts
```diff
@@ -522,20 +522,22 @@ export class TaleFilesComponent implements OnInit, OnChanges {
   moveElement(event: { element: FileElement; moveTo: FileElement }): void {
     const src = event.element;
     const dest = event.moveTo;
-    const params = { id: src._id, parentId: dest._id };
+    const params = { id: src._id, parentId: dest._id, parentType: ParentType.Folder, baseParentId: dest.baseParentId };
```
Xarthisius, line 525: These parameters only apply to folders (`parentId` and `parentType` in particular), yet they're passed as an argument to `this.itemService.itemUpdateItem()` below. `PUT /item/:id` will expect `folderId` as new target.

Xarthisius (changes requested):
Moving doesn't work for item/files right now. I highlighted the problem inline.

src/app/+run-tale/run-tale/tale-files/tale-files.component.ts
bodom0015: Whoops, had not noticed the difference in parameter name.. Good catch!

Will push a fix for this ASAP :+1:

Xarthisius (approved):
Works like a charm! With the addition of `virtual_resources` it's also possible to move in between `Home` and `Workspaces` as expected.

ThomasThelen (approved):

PR #106, opened 2020-10-21
fix: adjust Close Tale link based on accessLevel. See #103

Xarthisius:
## Problem
See #103. The same thing only with "Close" button.
  
## Approach
Copy and paste Mike's solution and pretend that I know what I'm doing.
  
## How to Test
See #103 `s/"Return to Dashboard"/"Close" button/g`


bodom0015 (approved):
Good catch! I totally forgot about the Close button lol

:+1: works as advertised

PR #113, opened 2020-12-10
Restyled login view/footer, removed footer elsewhere

### LINKED ISSUE #112: Update landing page and remove footer
We need to update the landing page (https://dashboard.wholetale.org) to a link to terms of service as well as several acknowledgements. While we're at it, we could update it to look more like wholetale.org.

Requirements:
* Add statement about terms of service under login button
* Add acknowledgement statement that includes NSF award, XSEDE, Jetstream and StataCorp
* Replace footer with partner logos (ala wholetale.org)
* Update landing page to look more like wholetale.org (i.e., background image, fonts, etc)

With these changes, we agreed that we could also drop the footer elsewhere in the UI.

I tried to do this myself but lack the skills, instead opting for a mockup:

https://docs.google.com/presentation/d/1uQ6R7gprhu8jH3UPFTnsNNezy9OhM0JzdGjxMqkj0JY/edit#slide=id.p

bodom0015:
## Problem
Footer takes up a lot of screen real estate, and does not very closely match the footer on wholetale.org

Fixes #112 

## Approach
* Adjust login view to previous include info from footer
* Adjust footer to more closely match wholetale.org, and to only display on the login view
* Add a line to the login view: `By accessing Whole Tale, you agree to our Terms of Use.` with a link to the terms
* Mention / link to affiliate universities
* Mention Jetstream / Stata license on login view (do these need links as well?)

Screen Shot 2020-12-10 at 1 06 44 PM

## How to Test
2. Navigate to the Login view
    * You should see that the styling on this page more closely matches that of wholetale.org
    * You should see that the footer on this page more closely matches that of wholetale.org
    * You should see that the login view now requires users to implicitly agree to the Terms of Use
    * You should see that the login view now mentions the Grant Number, Jetstream resources, and the Stata license (do these need links?)
    * NOTE: I found it difficult to change the background color on only the login view - if the subtle blue background feels odd, I can revisit to try and color it white, as it is in the mockup
3. Click on "Access WholeTale" and login to the WholeTale platform
    * You should see that the footer has been removed from all other views
4. Play around with other views to make sure that removing the footer didn't have any other weird side-effects elsewhere


Xarthisius:
I think I still get what you called a "phantom margin":

wt

The window with interact view has a fairly small width to show comparison with landing page, but it behaves the same way on a  full screen.

src/app/login/login.component.html
```diff
@@ -5,90 +5,62 @@
...
-        <div class="ui large header">
-          <p class="botmargin">Analyze/</p>
-          <p class="nomargin">Compute</p>
-        </div>
-      </div>
-
-      <div class="two wide column">
-      </div>
+  <p style="botmargin" style="color:black;">By accessing Whole Tale, you agree to our
+    <a href="https://docs.wholetale.org/tos" style="color:black"><u>Terms of Use.</u></a>
```
craig-willis, line 23: Period outside of link:
```suggestion
    <a href="https://docs.wholetale.org/tos" style="color:black"><u>Terms of Use</u></a>.
```

src/app/login/login.component.html
```diff
@@ -5,90 +5,62 @@
...
-          </div>
-          <a class="ui button" target="_blank" href="https://wholetale.readthedocs.io/en/stable/users_guide/sign_in.html">Signing In</a>
-        </div>
-      </div>
-    </div>
+  <div style="color:#666">
+    <p class="nomargin" style="margin-top: 60px; font-size: 14pt;">Whole Tale is supported by the National Science Foundation under Grant No. OAC-1541450.</p>
+    <p class="nomargin" style="font-size: 14pt;">Computational resources are provided by XSEDE through Jetstream.</p>
+    <p class="botmargin" style="font-size: 14pt;">Access to Stata is provided by StataCorp LLC.</p>
   </div>
```
craig-willis, line 30: I think black looks better here and a little smaller:

```suggestion
  <div style="color:black">
    <p class="nomargin" style="margin-top: 50px; font-size: 12pt;">Whole Tale is supported by the National Science Foundation under Grant No. OAC-1541450.</p>
    <p class="nomargin" style="font-size: 12pt;">Computational resources are provided by XSEDE through Jetstream.</p>
    <p class="botmargin" style="font-size: 12pt;">Access to Stata is provided by StataCorp LLC.</p>
  </div>
```

src/app/login/login.component.scss
```diff
@@ -15,17 +15,24 @@
   margin-top: 10vh;
 }
 
-.login.area {
-  background-image: url(../../assets/img/login-backdrop.png);
-  background-size: cover;
+.login.banner {
+  //background-image: url(../../assets/img/login-backdrop.png);
+  background-size: 100% 300px;
```
craig-willis, line 20: The background image aspect stretches when I resize the window. Cover seems to fix that, but you changed it for a likely good reason.
```suggestion
  background-size: cover;

```

craig-willis (approved):
Looks great. Thanks for your help.

> Mention Jetstream / Stata license on login view (do these need links as well?

I thought about linking to Jetstream and Stata site, but I don't think it's necessary. Mentioning should be sufficient.

@Xarthisius phantom footer is the only real issue I noted.  

A couple of super minor things:
* Background image aspect stretches when I resize the window.
* I'd take the period out of the "Terms of Use." link

Otherwise I've made a few take-or-leave-it styling suggestions, since my mockup was in a Google slide...

src/app/login/login.component.scss
bodom0015: Only changed it because I haven't worked much with background images... `cover` is likely still correct if it worked for you :+1:

bodom0015:
Good point.. I guess the Run view still has some hard-coded heights that need to be examined more closely.

I will take a look to see if I can get it to stretch to fit the new height there

bodom0015:
Phantom margins have been removed :tada:
This should also address @ThomasThelen's earlier point about the inconsistent scroll bar behavior

http://recordit.co/8lQqRq5jZq

Xarthisius:
I tested with 9b5df345ac5b0913881c8c2b9740289997dfcf3c and I've still got about 80-90px empty space at the bottom that's taken by an empty `<div>`. Tested it both on chrome and firefox

wt

bodom0015:
@Xarthisius good catch! I had not focused on testing with screens larger than the one on my Macbook.
While my current hardware options are limited, I've re-tested on a 1080p desktop monitor to reproduce the reported behavior, and have tried my best to make sure that this case does not include too much additional margin.

Unfortunately, I was unable to quickly find a bullet-proof solution here - notice the small margin and slight scroll in the Macbook case. For `.wt-panel-body`, `height:85%` extended too far for the laptop case, but `height:80%` left too much space for the monitor case. I've split the difference with `82%`, which seems to maximize space for smaller screens (without looking too broken), while still preserving the correct styling for larger screens.

This may require a deeper dive, if we find further margin faults under particular circumstances.


Desktop monitor:
Screen Shot 2020-12-17 at 1 34 47 PM

Macbook:
Screen Shot 2020-12-17 at 1 35 17 PM

bodom0015:
I should note that the problem **does** still exhibit in 4K, but the impact is minor as there is plenty of usable space to work with: 
Screen Shot 2020-12-17 at 1 48 32 PM

I think it should also be reasonably safe to assume that isn't going to be a common use case for this platform.

Xarthisius (approved):

ThomasThelen (approved):

PR #114, opened 2020-12-17
Avoid using placeholder in places that are not loading anything dynamically

Xarthisius:
## Problem

Default Semantic UI's `placeholder` class that creates a pulsating gradient really triggers me (in a bad way). 

## Approach

Use `message` class instead.

## How to Test

Behold!
wt


craig-willis (approved):

bodom0015 (approved):
LGTM :+1: Thank you! I agree that the `message` is much less distracting. I didn't realize that the `placeholder` class was meant for content that is still loading, but you are correct! I definitely should have read the docs a bit more closely :X

I also made a couple of other minor changes:
* fixed `.ui.placeholder` styles to instead apply to `.ui.message`
* removed some legacy styles in `theme.scss` that were messing with the bottom margin
* removed `.ui.message` border radius style to match the directory header

PR #126, opened 2021-01-21
fix: redirect from run-tale to catalog on error

### LINKED ISSUE #108: Invalid Instance ID doesn't redirect to 'Browse'
This was a failure from one of the regression tests in the dashboard testing document.

**Steps to Reproduce**

1. Open the dashboard
2. Navigate to a Tale's view page
3. Change the URL so that there's an invalid ID
4. Hit Enter
5. See that the notification stream gives an error and the page isn't redirected to 'Browse' (old dashboard behavior)

**Expected Results**

1. I don't see any errors and I'm brought to the browse page

**Actual Results**

1. See that the notification stream gives an error and the page isn't redirected to 'Browse' (old dashboard behavior)

bodom0015:
## Problem
The EmberJS dashboard used to redirect the user to the browse view when an invalid Tale ID was encountered. AngularJS dashboard needs this behavior as well.

Fixes #108 

## Approach
Redirect to "public" catalog if an error is encountered fetching the Tale in the Run Tale view.

## How to Test

3. Navigate directly to `/run/thisisnotarealid` in your browser
    * You should be redirected to the Public Tales catalog

### DIFF
--- a/src/app/+run-tale/run-tale/run-tale.component.ts
+++ b/src/app/+run-tale/run-tale/run-tale.component.ts
@@ -145,7 +145,7 @@ export class RunTaleComponent extends BaseComponent implements OnInit, OnChanges
                       .subscribe((tale: Tale) => {
         if (!tale) {
           this.logger.error("Tale is null, something went horribly wrong");
-
+          this.router.navigate(['public']);
           return;
         }
 
@@ -170,6 +170,7 @@ export class RunTaleComponent extends BaseComponent implements OnInit, OnChanges
         });
       }, err => {
         this.logger.error("Failed to fetch tale:", err);
+        this.router.navigate(['public']);
       });
     }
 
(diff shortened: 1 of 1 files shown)


Xarthisius (approved):

craig-willis (approved):
This works as expected. I'll note that the toast "HTTP failure response for https://girder.local.wholetale.org/api/v1/instance?taleId=thisisnotarealid: 400 OK" isn't particularly useful, but @Xarthisius pointed out that this is the case for all errors.

bodom0015:
I can remove the toast messages, if they might be less distracting... They seemed like a good idea at first, but I agree that it may not be helpful or professional to advertise the specific errors we might be hitting. Maybe it would be better to just replace this with a message that says something like this?
> An error was encountered. Please try again in a few moments, and contact an administrator if the error continues to occur.

craig-willis:
I'll open a separate issue for the error message discussion. I'd leave it as is for now.

ThomasThelen (approved):

PR #129, opened 2021-01-25
Update notification polling url after hitting acknowledge. Fixes #107

### LINKED ISSUE #107: Notifications reappear after acknowledge all
**Steps to Reproduce**
* Run a tale
* Notification window appears
* Wait for tale to start completely
* Click "Acknowledge all"
* Wait

This happens to me for every tale I build or run, every time.

**Expected Results**
Notification window does not reappear unless there is a new notification

**Actual Results**
Notification window appears displaying dismissed notifications

**NOTES**
* MacOS, Chrome 86.0.4240.75
* `lastRead` value is updated in local storage, but appears to be unused.
* Network indicates call to `https://girder.local.wholetale.org/api/v1/notification/stream?timeout=30&lastEventId=&r=8901482285531686`, but the Girder notification/stream API has only `timeout` and `since` parameters. I don't know where `lastEventId` and `r` are coming from.

Xarthisius:
## Problem

AFAICT problem lies entirely on the upstream package that were using. Once the EventSourcePolyfill is initialized there's no real way of changing the url, that contains `since` as a hardcoded parameter.

## Approach
 I made a couple of unrelated changes in this PR, but the only thing that matters is this:

https://github.com/whole-tale/ngx-dashboard/blob/be750721ca9049481b47277a66f5be421baad96e/src/app/layout/notification-stream/notification-stream.service.ts#L56

I have no idea if this can be done more cleanly, but it works for me...

## How to Test

Prerequisites: 

2. Login to the WholeTale Dashboard, preferably with Dev console open and identify the `GET /stream?since=`. Note `since` value
3. Open Notification panel and click on *Acknowledge All*
4. Wait and observe network tab in Dev console. Subsequent `GET /stream` calls should have `since` value updated (you may need to wait for more than one message, i.e. >60s)
Note: you can of course do some actions that trigger notifications in between steps 3 and 4 to check it really works ;-)

### DIFF
--- a/src/app/layout/notification-stream/notification-stream.service.ts
+++ b/src/app/layout/notification-stream/notification-stream.service.ts
@@ -14,6 +14,7 @@ class NotificationStreamService implements OnDestroy {
   static readonly Path = '/notification/stream';
   static readonly TimeoutMs = 30;
   static readonly IntervalDelayMs = 30000;
+  private _since: number = 0;
 
   interval: any;
 
@@ -24,7 +25,7 @@ class NotificationStreamService implements OnDestroy {
 
   ackAll() {
     const newSince = new Date().getTime() / 1000;
-    this.setSince(newSince);
+    this.since = newSince;
     //this.reconnect(true);
     //this.connect();
     this.openNotificationStream(false);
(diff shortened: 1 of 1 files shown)


craig-willis (approved):
This is working as expected for me.

bodom0015 (approved):
Good catch! I think at some point, `connect` or `reconnect` used to perform updating the url parameters, but I didn't think to update the internal source `url` directly.

LGTM :+1:

PR #130, opened 2021-01-25
fix: ensure that tale-interact handles iframe=False

### LINKED ISSUE #116: Interact page does not handle "iframe=false" case
I ran into this while testing a new environment that cannot be displayed in an iframe. 

**Steps to Reproduce**
* Modify `setup_girder.py` or use `girder-shell` in the `girder` container to set `iframe=False` for an image
* Create a tale using that image and run it

**Expected Results**
Per the old dashboard behavior (https://github.com/whole-tale/dashboard/blob/master/app/components/ui/tale-tab-interact/template.hbs#L1-L16) a message is displayed indicating that the interactive environment must be opened in a separate browser window.

**Actual Results**
No message is displayed and iframe support is assumed.

bodom0015:
## Problem
`tale-interact` tab may not properly handle cases when `iframe=False`

Fixes #116 

## Approach
Re-test this case and make sure styling is correct

Screen Shot 2021-01-25 at 3 47 46 PM

## How to Test
Prerequisites: at least on image with `iframe=False`:
  * You can use `GET/PUT /image/:id` on Swagger UI to toggle this easily

3. Create a new Tale from the image with `iframe=False`
4. Click Run Tale to navigate to Tale Interact
    * You should see a message indicating that this Tale cannot run in a n iframe and that it needs to run in a new window
    * You should be offered a button to open the Tale in a new window

### DIFF
--- a/src/app/+run-tale/run-tale/tale-interact/tale-interact.component.html
+++ b/src/app/+run-tale/run-tale/tale-interact/tale-interact.component.html
@@ -14,7 +14,13 @@
 
         <div *ngIf="!instance.iframe" class="ui message">
             <p>This Tale must be run in a separate window.</p>
-            <p><a class="ui button primary" [href]="instance.url" target="_blank">Click here to comply</a></p>
+            <p><a class="ui button primary"
+                  [href]="instance.url"
+                  [matTooltipShowDelay]="2000"
+                  matTooltip="Take a deep breath. Calm your mind. You know what is best. What is best is you comply. Compliance will be rewarded."
+                  target="_blank">
+              Click here to comply</a>
+            </p>
         </div>
       </div>
 
--- a/src/app/+run-tale/run-tale/tale-interact/tale-interact.component.html
+++ b/src/app/+run-tale/run-tale/tale-interact/tale-interact.component.html
@@ -13,14 +13,15 @@
         <iframe *ngIf="instance.iframe" height="100%" width="100%" [src]="instance.url | safe:'url'"></iframe>
 
         <div *ngIf="!instance.iframe" class="ui message">
-            This Tale must be run in a separate window.
-            <a class="ui button primary" [href]="instance.url" target="_blank">Click here to comply</a>
+            <p>This Tale must be run in a separate window.</p>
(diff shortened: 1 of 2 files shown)


src/app/+run-tale/run-tale/tale-interact/tale-interact.component.html
```diff
@@ -13,14 +13,15 @@ <h3>Tale is launching, please wait...</h3>
         <iframe *ngIf="instance.iframe" height="100%" width="100%" [src]="instance.url | safe:'url'"></iframe>
 
         <div *ngIf="!instance.iframe" class="ui message">
-            This Tale must be run in a separate window.
-            <a class="ui button primary" [href]="instance.url" target="_blank">Click here to comply</a>
+            <p>This Tale must be run in a separate window.</p>
+            <p><a class="ui button primary" [href]="instance.url" target="_blank">Click here to comply</a></p>
```
Xarthisius, line 17: I just can't pass this opportunity...
```suggestion
            <p><a class="ui button primary"
                  [href]="instance.url"
                  [matTooltipShowDelay]="2000"
                  matTooltip="Take a deep breath. Calm your mind. You know what is best. What is best is you comply. Compliance will be rewarded."
                  target="_blank">
              Click here to comply</a>
            </p>
```

craig-willis: Hmm.  Perhaps "Click here to open a new window" or simply "Click here" would suffice?

Xarthisius: No, cause that kills the easter egg... We are also discussing a page that won't be accessible to users, since all our images have `iframe=True`

craig-willis (approved):

ThomasThelen (approved):

src/app/+run-tale/run-tale/tale-interact/tale-interact.component.html
bodom0015: I do like Easter eggs like this :)

That said, if you don't get the reference then this is objectively creepy.

PR #134, opened 2021-01-27
Enable version restoring

Xarthisius:
## Problem

This PR enables "Restore" functionality in the version context menu.  Note that I'm merely calling a proper backend route, but I don't do anything with the state of the dashboard. In other words after restoring you need to refresh your browser (sorry!)

## How to Test

See https://github.com/whole-tale/wt_versioning/pull/15


bodom0015:
Basic restore functionality seems to work :+1: I added a few lines to pass a refresh signal up to the parent `run-tale.component`. This triggers a cascading refresh of the Tale data, collaborators, metadata, files, etc and automatically updates the view.

```
// TODO: Once they do this, how can the user get back to HEAD?
```
I noticed this `TODO` that @Xarthisius added that brings up an interesting point: we don't appear to track the currently-restored version anywhere.


For example:
1. Create a Tale: `version-test`
2. Create a Folder: `version1-files`
3. Create a Version: `version1`
4. Create another Folder: `version2-files`
5. Create another Version: `version2`
6. Check `GET /tale/:id`
7. Restore back to `version1`
8. Check `GET /tale/:id` again - notice that the diff of these two states is identical

This means that, currently, if a user refreshes the dashboard, we lose all context of which version they are on.

Is there perhaps a different API endpoint that we should be hitting, or some other way that we are tracking which version is currently checked-out on disk? I see `versionsRootId` and `runsRootId` on the Tale object, but these appear to point to all versions for the Tale and not just the current one.

Xarthisius:
> Is there perhaps a different API endpoint that we should be hitting, or some other way that we are tracking which version is currently checked-out on disk? I see `versionsRootId` and `runsRootId` on the Tale object, but these appear to point to all versions for the Tale and not just the current one.

There's a field on a Tale that I forgot to expose called `restoredFrom` which holds the id of the last version that was used to restored the current Tale. I'll fix that.

src/app/+run-tale/run-tale/tale-files/tale-files.component.html
```diff
@@ -20,7 +20,7 @@
         [path]="currentPath"
         [placeholderMessage]="placeholderMessage"
         [preventNavigation]="isNavActive('external_data')"
-        [readOnly]="tale._accessLevel <= AccessLevel.Read"
+        [readOnly]="tale._accessLevel <= AccessLevel.Read || tale.restoredFrom"
```
Xarthisius, line 23: Tale restored from a version shouldn't be read only. Think about it like `git checkout <tag>`.

bodom0015: Whoops, I misunderstood that part.. 🤔 

I'll remove the read-only wording and behavior

craig-willis (approved):
This works for me.  I think we should remove the code related to restoring the head version, since that's not something we support.

bodom0015 (approved):
LGTM :+1: this is an oddly intuitive workflow... I had questions, but playing around with the feature answered them for me. I'm not sure if this is a good thing or a bad thing.

My test steps:
1. Create a Tale, go to Run
2. Create a version called `init`
3. Attach a dataset to the Tale
4. Create a version called `added dataset`
5. Restore to `init` version
6. Attach a different dataset to the Tale
7. Create another version called `different dataset`
8. Restore to `added dataset`, view External Data
9. Restore to `init`, view External Data
10. Restore to `different dataset`, view External Data

PR #136, opened 2021-01-28
fix: remove obsolete Workspace service

Xarthisius:
## Problem

Drops dead code related to now obsoleted backend path.

## How to Test

3. Play around, nothing should be broken.


bodom0015 (approved):
LGTM :+1: Thank you!

PR #138, opened 2021-02-11
feat: support allowRename, fix race condition

### LINKED ISSUE #117: Version name validation race condition on slow connection
**Steps to Reproduce**
1. Navigate to a Tale that you own
2. Create a version with a specific name (e.g. `my version`)
3. Open browser developer console to Networks tab and throttle your connection to "Slow 3G"
4. Type and confirm very quickly to create the same version again: `my version`

**Expected Results**
One of the following should occur:
1. UI should block modal confirmation until after input is fully validated
2. UI should not validate prior to submission and API should simply append `(1)` or similar to the end of the conflicting version name
3. UI should not validate prior to submission and should reopen the modal in case of error - Noting that this approach may not work well with `MatDialog`

Note that the `/version/exists` endpoint is no longer needed with option 2 or 3

**Actual Results**
User's input (which has not yet completed validation) is submitted to the server, resulting in an error from the POST endpoint complaining that the name is already being used for another version

bodom0015:
## Problem
Race condition when validating name allows user to submit invalid data. They get an error popup, but we should prevent this behavior.

Fixes #117 

Based on https://github.com/whole-tale/wt_versioning/pull/18

## Approach
* For "Create" case: support `allowRename` during CreateVersion - this will guarantee unique version names by appending `(1)` or similar to resolve conflicts
* For "Rename" case: block submission of the form until name has been fully validated

## How to Test
Prerequisites: running https://github.com/whole-tale/wt_versioning/pull/18, at least one Tale created

### Test Setup
3. Choose a Tale that you own and click "View"
4. Expand the Tale Versions Panel on the right side

### Test "Create" Case
1. Click on "Save New Version"
    * You should see the create-rename-version-modal appear
2. Enter `name` and submit (make sure `Force` is checked)
    * You should see that a new version called `name` appear in the list
3. Repeat steps 2 and 3
    * You should see that a new version called `name (1)` appear in the list

### Test "Rename" Case
1. Refresh the page
2. Expand the Tale Versions Panel on the right side
3. Expand the dropdown beside the Version called `name (1)` and choose "Rename"
    * You should see the create-rename-version-modal appear
4. Open the developer console to the Networks tab, expand the dropdown that says `Online`, and throttle your connection to a slower speed
5. Enter `name` in the modal
    * You should see the "Save" button is disabled while we validate the name you've entered
6. Wait for validation to complete
    * You should see the "Save" button is still disabled
    * You should see a red banner appear notifying you that the name is invalid

### DIFF
--- a/src/app/+run-tale/run-tale/modals/create-rename-version-dialog/create-rename-version-dialog.component.html
+++ b/src/app/+run-tale/run-tale/modals/create-rename-version-dialog/create-rename-version-dialog.component.html
@@ -17,7 +17,7 @@
                     <div class="sixteen wide column">
                         <div class="field" [ngClass]="{ 'error': nameAlreadyExists }">
                             <!-- Require new name in "Rename" mode -->
-                            <input *ngIf="data.mode==='create'" type="text" name="name" id="name" [(ngModel)]="name" (ngModelChange)="doesNameExist($event)" />
+                            <input *ngIf="data.mode==='create'" type="text" name="name" id="name" [(ngModel)]="name" />
                             <input *ngIf="data.mode!=='create'" type="text" name="name" id="name" [(ngModel)]="name" required (ngModelChange)="doesNameExist($event)" />
                         </div>
                     </div>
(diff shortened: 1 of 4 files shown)


craig-willis (approved):
This works for me.

Xarthisius (approved):

ThomasThelen (approved):

PR #140, opened 2021-02-20
Export Versions

ThomasThelen:
A small code change for adding a new row in the context menu for versions. This _should_ be the extent of the user interface changes for exporting versions.

To Do (WIP Tag) : As of this PR it's possible to test that the manifest has the correct version record, but I think it makes sense to do a more full test with a girder_wholetale PR for exporting particular versions & creating a version when one doesn't exist.

### Testing
The gist of the testing is to make sure that particular versions of a Tale are exporting. To test, create a few different versions of a Tale, each with different kinds of data (workspace/remote data).

1. While running `make rebuild_dashboard`, check the output to see that there aren't any new linter warnigns.
1. Create a new Tale
2. Add data to the workspace & remote data
3. Create a version
4. Delete or add more data to each of the folders above
5. Create a new version
6. In the context menu for the first version, select 'export'
7. Confirm that the exported Tale matches the first version
8. In the context menu for the second version, select 'export'
9. Confirm that the exported Tale matches the second version

### DIFF
--- a/src/app/+run-tale/run-tale/tale-versions-panel/tale-versions-panel.component.html
+++ b/src/app/+run-tale/run-tale/tale-versions-panel/tale-versions-panel.component.html
@@ -49,6 +49,7 @@
 
                 <div class="item" (click)="renameVersion(version)" *ngIf="tale._accessLevel >= AccessLevel.Write">Rename</div>
                 <div class="item" (click)="deleteVersion(version)" *ngIf="tale._accessLevel >= AccessLevel.Write">Delete</div>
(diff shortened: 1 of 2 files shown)


craig-willis (approved):

bodom0015 (approved):
LGTM :+1: zip file is downloaded as expected

it might make sense to change the resulting zip file name to be something like `TALENAME-VERSIONNAME.zip`, but for simplicity's sake I definitely understand the decision to use the versionId as the zip filename.

Xarthisius (approved):

PR #146, opened 2021-03-21
fix: allow to pass config via env vars. Fixes #145

### LINKED ISSUE #145: Hardcoded URLs
Deploying the new dashboard on staging, I noticed that the API root URL and DataONE repository URL are hardcoded in the source.  For the old dashboard, the [entrypoint 
 script](https://github.com/whole-tale/dashboard/blob/master/docker-entrypoint.sh) handled replacement of these values.

`rootUrl`:
https://github.com/whole-tale/ngx-dashboard/blob/0f90e6f0574b498eeda54099515bb5eb472468d5/src/app/api/api-configuration.ts#L10-L12

`repositoryBaseUrl`:
https://github.com/whole-tale/ngx-dashboard/blob/0f90e6f0574b498eeda54099515bb5eb472468d5/src/app/%2Brun-tale/run-tale/modals/register-data-dialog/register-data-dialog.component.ts#L22

As before we can use `sed` to do variable replacement in the entrypoint, but maybe there's a better way with Angular?

Xarthisius:
## Problem

Some urls / variables needs to be configurable via env vars.

## Approach
https://pumpingco.de/blog/environment-variables-angular-docker/

## How to Test

1. Deploy with deploy-dev (note: GIRDER_URL_API should point to `https://girder.local.wholetale.org/api/v1` whereas currently in deploy-dev@master it points to `https://girder.local.wholetale.org`. Unless you change it, you'll get a dashboard's blue screen of death. That's also how you can test this :) )
2. (Optionally) Deploy on .stage.local.wholetale.org (build and push image, or use wholetale/ngx-dashboard:env)


craig-willis (approved):
lgtm

Xarthisius:
I'm a little bit worried that this fails `ci/dockercloud-stage`, however when I click that link I get 404....

Xarthisius:
I completely rewrote this PR using  existing `WindowService` instead. Unfortunately I couldn't use it `src/app/api/api-configuration.ts` which required additional magic and I won't pretend I understand it fully...

bodom0015 (approved):
Apparently, "blue screen of death" translates roughly to "CORS errors" 😆 LGTM

PR #158, opened 2021-04-06
Double refresh on delete

craig-willis:
**Problem**:
With the new event handling implementation, deleting a tale from the catalog results in a double refresh.

**Approach**:
Remove the refresh from the `openDeleteTaleModal` and rely on the tale removed event handling instead.

**Test**:
1. Create a tale
2. Return to dashboard
3. Delete the tale and confirm that the view is only refreshed once

### DIFF
--- a/src/app/+tale-catalog/tale-catalog/components/my-tales/my-tales.component.html
+++ b/src/app/+tale-catalog/tale-catalog/components/my-tales/my-tales.component.html
@@ -1,6 +1,6 @@
 
 <!-- My Tales -->
-<div *ngIf="(tales$ | async) as taleList">
+<div *ngIf="tales.length">
 
     <!-- Search Tales -->
     <div class="row">
@@ -13,14 +13,14 @@
     </div>
 
   <!-- Running Tales -->
-  <app-running-tales [instances]="instances" [creators]="creators" [taleList]="taleList"></app-running-tales>
+  <app-running-tales [instances]="instances" [creators]="creators" [taleList]="tales"></app-running-tales>
 
-  <div class="row" style="margin-top:40px;margin-bottom:30px;" *ngIf="(taleList | myTales | stopped:instances) as allStoppedTales">
-    <div class="sixteen wide column" *ngIf="(taleList | searchTales:searchQuery:creators | myTales | stopped:instances) as filteredStoppedTales">
+  <div class="row" style="margin-top:40px;margin-bottom:30px;" *ngIf="(tales | myTales | stopped:instances) as allStoppedTales">
+    <div class="sixteen wide column" *ngIf="(tales | searchTales:searchQuery:creators | myTales | stopped:instances) as filteredStoppedTales">
 
       <!-- Placeholder Message -->
       <div class="ui message" *ngIf="!allStoppedTales.length">
-          You haven't created any Tales yet. Select the "Create New Tale" button above 
+          You haven't created any Tales yet. Select the "Create New Tale" button above
           or <a [routerLink]="'/public'">explore Tales created by other users</a>.
       </div>
 
--- a/src/app/+tale-catalog/tale-catalog/components/public-tales/public-tales.component.html
+++ b/src/app/+tale-catalog/tale-catalog/components/public-tales/public-tales.component.html
@@ -1,6 +1,6 @@
 
 <!-- Public Tales -->
-<div *ngIf="(tales$ | async) as taleList">
+<div *ngIf="tales.length">
 
     <!-- Search Tales -->
     <div class="row">
@@ -13,10 +13,10 @@
     </div>
 
   <!-- Running Tales -->
-  <app-running-tales [instances]="instances" [creators]="creators" [taleList]="taleList"></app-running-tales>
+  <app-running-tales [instances]="instances" [creators]="creators" [taleList]="tales"></app-running-tales>
 
-  <div class="row" style="margin-top:40px;margin-bottom:30px;" *ngIf="(taleList | publicTales) as allPublicTales">
-    <div class="sixteen wide column" *ngIf="(taleList | searchTales:searchQuery:creators | publicTales) as filteredPublicTales">
+  <div class="row" style="margin-top:40px;margin-bottom:30px;" *ngIf="(tales | publicTales) as allPublicTales">
+    <div class="sixteen wide column" *ngIf="(tales | searchTales:searchQuery:creators | publicTales) as filteredPublicTales">
 
       <div class="ui message" *ngIf="!allPublicTales.length">
           No Public Tales have been shared.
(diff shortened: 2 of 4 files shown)


Xarthisius:
LGTM. Just noting that prior to this PR there are 2 updates: one immediate and one slightly delayed. This PR removes the immediate one, which may or may not be desired.

Xarthisius (approved):

bodom0015 (approved):
LGTM :+1: as long as the event is triggering the refresh, this can be removed and was likely an oversight in adding the new event handling

Noting @Xarthisius' comment that the first refresh might pick up a "Tale deleting" state, if such a state exists. Perhaps if the server is cleaning up resources in the background (e.g. a large Tale workspace i being removed)... but if this is not currently supported, then it is likely not something that we need to consider.

craig-willis:
In hindsight, this might not be the best approach.

The lines I've removed are re-loading the tale list and refreshing the view immediately after the delete request returns.  The event is received only after the delete is completed, which as noted could take a while for a large workspace. This means that there could be a phantom tale in the list after delete. The event approach was really intended for other users anyway.

Maybe it would be better to do a targeted removal from the view of the specific tale instead of a full refresh?  Then we could both remove after the delete API is called and the event would be a no-op for the deleter.

craig-willis:
I'm out of my depths on this.  I've tried to remove the deleted tale from `tales$` via `splice`, but since `tales$` is "observable" it doesn't support it. To remove an item from an observable array seems to require use of Subjects, which is beyond me.

bodom0015:
@craig-willis you were right to be confused - that pattern with `tales$` was a mess. I've removed that altogether, and I think that simplifies things a lot.

The Tale now appears to deleted immediately. Let me know if this resolves what you were seeing :+1:

Xarthisius:
@bodom0015 now I see something totally weird:

### How to reproduce?
1. Create a Tale named `Tale #1`
2. Create a Tale named `Tale #2`
3. Navigate to dashboard / mine
4. Remove `Tale #2`

### Expected result
`Tale #2` disappears

### Actual result
`Tale #1` disappears, but if you wait a few seconds remaining `Tale #2` gets renamed to `Tale #1` (after 2nd refresh() I guess).

bodom0015:
Whoops, good call! I somehow forgot that `.splice()` modifies the array in-place, and the return value is the array value that was removed.

I believe that this may actually be the reason why splice doesn't trigger the change-detection, which may only trigger if our variable reference changes - as a workaround, we can use `.filter()` to overwrite the entire list with a new list containing everything except what we're removing, triggering the change-detection as expected.

bodom0015:
Build failed, but I suspect that it's just an intermittent failure (HTTP 500 during `yarn install`.
Re-running the build to find out.

Build logs, for reference:
```
Step 4/13 : RUN $YARN install --network-timeout=360000 && $YARN cache clean
---> Running in ba80fdca6229
yarn install v1.22.4
[1/4] Resolving packages...
[2/4] Fetching packages...
[91mwarning Pattern ["colors@latest"] is trying to unpack in the same destination "/usr/local/share/.cache/yarn/v6/npm-colors-1.4.0-c50491479d4c1bdaed2c9ced32cf7c7dc2360f78-integrity/node_modules/colors" as pattern ["colors@^1.4.0"]. This could result in non-deterministic behavior, skipping.
[0m
[91mwarning Pattern ["object-assign@latest"] is trying to unpack in the same destination "/usr/local/share/.cache/yarn/v6/npm-object-assign-4.1.1-2109adc7965887cfc05cbbd442cac8bfbb360863-integrity/node_modules/object-assign" as pattern ["object-assign@^4.0.1","object-assign@^4.1.1","object-assign@^4.0.1","object-assign@^4.1.0","object-assign@^4.1.1","object-assign@^4","object-assign@^4.1.0","object-assign@^4.1.0","object-assign@^4.0.1","object-assign@^4.1.1"]. This could result in non-deterministic behavior, skipping.
warning Pattern ["send@latest"] is trying to unpack in the same destination "/usr/local/share/.cache/yarn/v6/npm-send-0.17.1-c1d8b059f7900f7466dd4938bdc44e11ddb376c8-integrity/node_modules/send" as pattern ["send@0.17.1","send@0.17.1"]. This could result in non-deterministic behavior, skipping.
[0m
[91mwarning file-loader@2.0.0: Invalid bin field for "file-loader".
[0m
[91mwarning mini-css-extract-plugin@0.4.4: Invalid bin field for "mini-css-extract-plugin".
[0m
[91merror An unexpected error occurred: "https://registry.yarnpkg.com/indexes-of/-/indexes-of-1.0.1.tgz: Request failed \"500 Internal Server Error\"".
[0m
info If you think this is a bug, please open a bug report with the information provided in "/srv/app/yarn-error.log".
info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.
Removing intermediate container ba80fdca6229
The command '/bin/sh -c $YARN install --network-timeout=360000 && $YARN cache clean' returned a non-zero code: 1
```

bodom0015:
Build failed, but I suspect that it's just an intermittent failure (HTTP 500 during `yarn install`.
Re-running the build to find out.

Build logs, for reference:
```
Step 4/13 : RUN $YARN install --network-timeout=360000 && $YARN cache clean
---> Running in ba80fdca6229
yarn install v1.22.4
[1/4] Resolving packages...
[2/4] Fetching packages...
[91mwarning Pattern ["colors@latest"] is trying to unpack in the same destination "/usr/local/share/.cache/yarn/v6/npm-colors-1.4.0-c50491479d4c1bdaed2c9ced32cf7c7dc2360f78-integrity/node_modules/colors" as pattern ["colors@^1.4.0"]. This could result in non-deterministic behavior, skipping.
[0m
[91mwarning Pattern ["object-assign@latest"] is trying to unpack in the same destination "/usr/local/share/.cache/yarn/v6/npm-object-assign-4.1.1-2109adc7965887cfc05cbbd442cac8bfbb360863-integrity/node_modules/object-assign" as pattern ["object-assign@^4.0.1","object-assign@^4.1.1","object-assign@^4.0.1","object-assign@^4.1.0","object-assign@^4.1.1","object-assign@^4","object-assign@^4.1.0","object-assign@^4.1.0","object-assign@^4.0.1","object-assign@^4.1.1"]. This could result in non-deterministic behavior, skipping.
warning Pattern ["send@latest"] is trying to unpack in the same destination "/usr/local/share/.cache/yarn/v6/npm-send-0.17.1-c1d8b059f7900f7466dd4938bdc44e11ddb376c8-integrity/node_modules/send" as pattern ["send@0.17.1","send@0.17.1"]. This could result in non-deterministic behavior, skipping.
[0m
[91mwarning file-loader@2.0.0: Invalid bin field for "file-loader".
[0m
[91mwarning mini-css-extract-plugin@0.4.4: Invalid bin field for "mini-css-extract-plugin".
[0m
[91merror An unexpected error occurred: "https://registry.yarnpkg.com/indexes-of/-/indexes-of-1.0.1.tgz: Request failed \"500 Internal Server Error\"".
[0m
info If you think this is a bug, please open a bug report with the information provided in "/srv/app/yarn-error.log".
info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.
Removing intermediate container ba80fdca6229
The command '/bin/sh -c $YARN install --network-timeout=360000 && $YARN cache clean' returned a non-zero code: 1
```

Xarthisius (approved):

PR #159, opened 2021-04-06
Add "report a problem" button to the navbar

bodom0015:
## Problem
"Report a Problem" button went away when the footer was removed/hidden.

## Approach
* Add this functionality back in the navbar
* Add tooltips to pulled-right navbar links

## How to Test
    * You should see the "bug" icon in the navbar
3. Hover over the "bug" icon in the navbar (and Users' Guide / Notification buttons)
    * You should see a tooltip appear after a few seconds reading "Report a Problem"
4. Click the "bug" icon
    * You should see https://github.com/whole-tale/whole-tale/issues open in a new tab

### DIFF
--- a/src/app/layout/header.component.html
+++ b/src/app/layout/header.component.html
@@ -5,10 +5,13 @@
     <!-- <a class="item" [ngClass]="{ 'active':currentRoute==='/datasets' }" routerLink="/datasets" routerLinkActive="active">Data Catalog</a> -->
     <!-- <a class="item" [ngClass]="{ 'active':currentRoute==='/environments' }" routerLink="/environments" routerLinkActive="active">Compute Environments</a> -->
     <div class="right menu">
-      <a class="item" href="https://wholetale.readthedocs.io/en/stable/users_guide/" target="_blank">
+      <a class="item" href="https://wholetale.readthedocs.io/en/stable/users_guide/" title="Users' Guide" target="_blank">
         <i class="info circle white icon"></i>
       </a>
-      <a class="item" (click)="toggleNotificationStream()">
+      <a class="item" href="https://github.com/whole-tale/whole-tale/issues" title="Report a Problem" target="_blank">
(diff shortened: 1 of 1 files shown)


Xarthisius:
LGTM! How about we point it to https://github.com/whole-tale/whole-tale/issues/new and save potential reporter one click of a mouse?

Xarthisius (approved):

craig-willis (approved):

bodom0015:
@Xarthisius this was my initial thought as well, but I think that putting them at the list of issues might encourage them to first search for a related issue. It also provides a nice bit of context as to what they're looking at and what goes into a ticket.

I don't think that repo currently offers a template for Bug Reports, but that might be a good compromise. If the need arises, we can always more issue templates/types there that can help guide the user through the ticket creation process, similar to what we have set up with https://github.com/whole-tale/wt-design-docs/issues/new/choose.

bodom0015:
> @Xarthisius this was my initial thought as well, but I think that putting them at the list of issues provides a nice bit of context as to what they're looking at and what goes into a ticket. Optimistically, it might even encourage them to first search for a related issue before filing.

Fixed, I guess?

We don't have many users filing issues right now, let alone duplicates, so probably a moot point anyways. I think that linking to "New Issue" will force them to login (if they don't use GitHub very often), after which they may have forgotten why they were even there in the first place. I don't think we're losing anything by forcing another mouse click, but we can certainly agree that the GitHub UI is not very intuitive to the uninitiated. I'm open to compromising by making that mouse click actually mean something.

For example, if the `/choose` view offers several different types of Bug Reports or steps for different sets of resolving common issues, this would be more beneficial and would offer even more context than dropping the user at a list of issues and expecting them to find the "New Issue" button. They would still need to click, but with the UI being so focused at least they would know where to click and would be empowered to do so.

Does this seem like a reasonable compromise?

PR #164, opened 2021-04-14
fix: make notification ack button stand out

Xarthisius:
## Problem

Fixes whole-tale/whole-tale#92

## Approach
Make it green (positive) button. I can be talked out of "fluid" in favor of "floated right".

## How to Test

1. Deploy
2. Click on notification panel and see how it looks and feels without any message
3. Create a Tale and build image
4. See how notification panel looks and feels.

### DIFF
--- a/src/app/layout/notification-stream/notification-stream.component.html
+++ b/src/app/layout/notification-stream/notification-stream.component.html
@@ -51,7 +51,7 @@
     </tr>
   </table>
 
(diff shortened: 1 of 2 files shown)


craig-willis (approved):
The `fluid` button is a bit overwhelming to me.  I'd vote for float (both right and left seem to align left for me) and possibly shrinking the font size. 

Screen Shot 2021-04-14 at 2 40 59 PM

bodom0015 (approved):
LGTM :+1:

PR #165, opened 2021-04-14
Add configurable documentation URL

### LINKED ISSUE #163: Configurable documentation root
When opening the staging environment for external feedback, it would be helpful to have the documentation point to latest instead of stable, ideally configurable via env var.

craig-willis:
Fixes #163 

## Problem
Documentation URLs are hardcoded, pointing to stable

## Approach
Add new env var `RTD_URL` that is used as a base URL in the creation of documentation links.

## How to Test

Deploy with https://github.com/whole-tale/deploy-dev/pull/48

2. Goto https://dashboard.local.wholetale.org/login, confirm TOS URL points to https://wholetale.readthedocs.io/en/latest/tos
3. Goto https://girder.local.wholetale.org/api/v1/integration/dataverse?datasetPid=doi%3A10.7910%2FDVN%2F3MJ7IR&siteUrl=https%3A%2F%2Fdataverse.harvard.edu, confirm "Why would I do this" points to latest
3. Check header "i", Tale > Read the docs, Publish, register data

### DIFF
--- a/src/app/+run-tale/run-tale/modals/publish-tale-dialog/publish-tale-dialog.component.html
+++ b/src/app/+run-tale/run-tale/modals/publish-tale-dialog/publish-tale-dialog.component.html
@@ -11,7 +11,7 @@
   </p>
 
   <p>For more information on how to publish and cite your Tale, please consult the
-    <a href="https://wholetale.readthedocs.io/en/stable/users_guide/publishing.html" target="_blank">
+    <a [href]="docUrl" target="_blank">
       Publishing Guide
     </a>.
   </p>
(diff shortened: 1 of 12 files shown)


Xarthisius (approved):

Xarthisius (approved):

bodom0015 (commented):
LGTM :+1: I see no problem with parameterizing this, but also noting that I'm not clear on the motivation here. Is this to facilitate testing the new docs match up properly (e.g. only on stage)?

Shortly after release, I would imagine that `latest` will get merged into `stable` - isn't the `stable` documentation likely where we should be linking in the long-term?

My thinking is that `latest` could contain unreleased features or unvetted documentation and may not match what the user is actually seeing.

craig-willis:
> Is this to facilitate testing the new docs match up properly (e.g. only on stage)?
Yes, and for external reviewers of the staging deployment to be able to see relevant documentation.

Ideally the RTD documentation would align with each environment -- dev=latest, staging=rc, prod=stable.

bodom0015 (approved):
That makes sense :+1: LGTM

PR #166, opened 2021-04-16
Added original/copy indicator on public tales

craig-willis:
## Problem

In the quickstart documentation, we instruct users to copy-on-launch the LIGO tale. However, on production there are now 3 public LIGO tales -- which one should they use?

## Approach
This PR adds the original/copy indicator to public tales

Screen Shot 2021-04-16 at 3 18 17 PM

## How to Test
* Deploy this PR
* Register LIGO tale
* Copy-on-launch LIGO tale and set to public
* Goto Public Tales tab and confirm that copy/original display


bodom0015 (approved):
LGTM :+1: works as advertised

PR #169, opened 2021-04-26
Remove padding around iframe

### LINKED ISSUE #168: MATLAB Desktop sized to fit iframe with no scroll
**Problem**
When starting MATLAB Desktop on stage, I noticed that the IDE auto-fits a very small iframe and is basically unusable without hitting the pop-out button. I don't know if this is an artifact of the MATLAB IDE or how we size the iframe in angular or a combination of both.

Screen Shot 2021-04-23 at 3 46 09 PM

**Test case**
* https://dashboard.stage.wholetale.org
* Create tale "MATLAB Desktop" from Git repo https://github.com/whole-tale/matlab-example
* Select environment MATLAB (Desktop) [R2020b]
* Run the tale

**Expected results**:
MATLAB IDE is available in iframe with usable dimensions and I can scroll to see the full IDE

**Actual results**:
The MATLAB IDE is sized to fit the visible iframe and is unusable. Of course I can hit the pop-out button which works fine.

bodom0015:
## Problem
There is a bunch of extra padding around the `iframe` on Run Tale > Interact. This makes it difficult to use the view within the `iframe`, as there is not much space to see.

Fixes #168 ?

## Approach
* Remove as much padding as possible from around the iframe without disturbing other parts of the UI.
* Add the same padding values back to Run Tale > Metadata

Screen Shot 2021-04-26 at 3 49 57 PM

Note that the bottom of the panel is a special case - because of the unique scrolling behavior and how painful it is/was to tweak those margins across all views. If we are ok with the panel stretching off the bottom of the page, then we can revisit this.

Also, please ignore the broken/fake login screen here (WT hates that I use `localhost:4200` for development and throws security errors, as mentioned in the dev meeting)

## How to Test
Prerequisites: at least one Tale running

3. Navigate to Run Tale > Interact for a running Tale
    * You should see the iframe now appears flush against the tab bar at the top, as well as the panel borders on the left and right

### DIFF
--- a/src/app/+run-tale/run-tale/run-tale.component.html
+++ b/src/app/+run-tale/run-tale/run-tale.component.html
@@ -1,5 +1,5 @@
 <div fxFlex fxLayout="row" fxLayoutAlign="space-between stretch" [@routeAnimation]>
-    <div class="ui stretched stackable grid panel-container" *ngIf="tale">
+    <div class="ui stackable grid panel-container" *ngIf="tale">
         <div class="row">
             <div class="sixteen wide column">
 
--- a/src/app/+run-tale/run-tale/run-tale.component.scss
+++ b/src/app/+run-tale/run-tale/run-tale.component.scss
@@ -157,13 +157,19 @@ app-tale-versions-panel {
   height: 100%;
 }
 
+.wt_panel_body > .ui.grid,
+.wt_panel_body > .ui.grid > .row,
+.wt_panel_body > .ui.grid > .row > .column {
+  margin: 0;
+  padding: 0;
+}
+
 .ui.grid {
   height: 100%;
 }
 
 .wt-panel .wt-panel-body {
-  padding: 2rem 2rem 0 2rem;
-  height: 82%;
+  height: 88%;
 }
 .wt-panel .wt-panel-body.bg-grey,
 .wt-panel .wt-panel-header.bg-grey {
--- a/src/app/+run-tale/run-tale/run-tale.component.scss
+++ b/src/app/+run-tale/run-tale/run-tale.component.scss
@@ -14,6 +14,7 @@
 
 :host {
   min-height: 95%;
+  margin: 12pt 0 0 0;
 }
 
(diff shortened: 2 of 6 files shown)


craig-willis:
Here's what it looks like with JupyterLab (for comparison to https://github.com/whole-tale/ngx-dashboard/issues/168#issuecomment-827020604): 

Screen Shot 2021-04-26 at 5 09 51 PM

craig-willis (approved):
This is certainly better and even has more space for the iframe than v0.9 (and fixes https://github.com/whole-tale/ngx-dashboard/issues/168 afaic). 

The double-scroll doesn't bother me and makes me wonder (again) if we can make the iframed area larger than the visible area -- i.e., set some reasonable minimum height that takes advantage of the browser scroll. 

I don't fully understand your comment about the bottom of the panel.  Aside from visual balance, I don't see any value in the extra space there.

bodom0015:
@craig-willis I 100% agree, but it was part of the original design. We can deviate from the design, but that's why I had stuck with it thus far.

The problem is that our whole view appears like it's in a panel - this begs the a question of what happens to the bottom of the panel when the content is too large. We mix the expected behavior here in different places in the UI. For most views, the height of the content is capped and causes an inner scroll to occur. In the Run > Metadata view (edit mode), however, the height is not capped and allows for a more natural scroll. If we prefer this behavior over the current desire to display the full panel all of the time.

Revisiting this design might also prevent a lot of the `max-height` problems that we had previously as well

bodom0015:
I've reworked this a bit to remove a bit more padding around the panel... I think the only way to get more space out of this view would be to remove the panel altogether.

Let me know what you think!

Screen Shot 2021-04-26 at 11 56 41 PM
Screen Shot 2021-04-26 at 11 56 31 PM
Screen Shot 2021-04-26 at 11 56 54 PM
Screen Shot 2021-04-26 at 11 57 04 PM

Xarthisius (approved):

PR #171, opened 2021-05-03
Don't spawn Tale with AiWT

bodom0015:
## Problem
Tale is launched after importing with AiWT.

## Approach
Pass `spawn=false` to prevent Tale from being automatically launched.

## How to Test
3. Navigate to https://girder.stage.wholetale.org/api/v1/integration/dataverse?datasetPid=doi%3A10.7910%2FDVN%2F3MJ7IR&siteUrl=https%3A%2F%2Fdataverse.harvard.edu
    * You should be redirected to the dashboard
    * You should see the Create Tale modal appear
    * You should see that "Create New Tale" is not enabled at the bottom
4. Enter a title and select an Environment
    * You should see now that "Create New Tale" is enabled
5. Click "Create New Tale"
    * You should be redirected to the Run view
    * You should **not** see the Tale automatically launch


craig-willis (approved):
Looks good.  While testing I noticed https://github.com/whole-tale/girder_wholetale/issues/476 and will try to fix it.

PR #175, opened 2021-05-10
fix: make 'Get token from url' for API keys dynamical

Xarthisius:
## Problem

`docs_href` in connect api_key account was hardcoded, but it would be nice if it took `resource_server` into account.

## Approach
Solves the problem by using a template string. Only show href if resource_server is selected.

## How to Test

1. Checkout this branch locally, rebuild the dashboard (needs https://github.com/whole-tale/girder_wholetale/pull/479)
3. Go to settings page
4. Select Dataone.
5. Select one of the CNs.
6. Follow "Get from DataONE"
7. Confirm you get a token.

### DIFF
--- a/src/app/+user-settings/modals/connect-apikey-modal/connect-apikey-modal.component.html
+++ b/src/app/+user-settings/modals/connect-apikey-modal/connect-apikey-modal.component.html
@@ -23,7 +23,7 @@
           </div>
           <div class="field">
             <label>API Token
-                <a *ngIf="data.provider.docs_href" href="{{ data.provider.docs_href }}" style="float:right; font-weight:normal" target="_blank">Get from {{ data.provider.fullName }} <i class="external square alternate icon"></i></a>
+              <a *ngIf="newApiKey.resource_server" href="{{ getToken() }}" style="float:right; font-weight:normal" target="_blank">Get from {{ data.provider.fullName }} <i class="external square alternate icon"></i></a>
             </label>
             <input type="text" name="API Token" placeholder="Enter an API key" [(ngModel)]="newApiKey.access_token" />
           </div>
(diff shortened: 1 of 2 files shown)


src/app/+user-settings/modals/connect-apikey-modal/connect-apikey-modal.component.ts
```diff
@@ -66,4 +66,8 @@ export class ConnectApiKeyModalComponent extends BaseComponent implements OnInit
   trackByTarget(index: number, target: string): string {
       return target;
   }
+
+  getToken(): string {
```
bodom0015, line 70:
```suggestion
  getTokenUrl(): string {
```

src/app/+user-settings/modals/connect-apikey-modal/connect-apikey-modal.component.html
```diff
@@ -23,7 +23,7 @@ <h1 mat-dialog-title>
           </div>
           <div class="field">
             <label>API Token
-                <a *ngIf="data.provider.docs_href" href="{{ data.provider.docs_href }}" style="float:right; font-weight:normal" target="_blank">Get from {{ data.provider.fullName }} <i class="external square alternate icon"></i></a>
+              <a *ngIf="newApiKey.resource_server" href="{{ getToken() }}" style="float:right; font-weight:normal" target="_blank">Get from {{ data.provider.fullName }} <i class="external square alternate icon"></i></a>
```
bodom0015, line 26:
```suggestion
              <a *ngIf="newApiKey.resource_server" href="{{ getTokenUrl() }}" style="float:right; font-weight:normal" target="_blank">Get from {{ data.provider.fullName }} <i class="external square alternate icon"></i></a>
```

bodom0015 (approved):
LGTM :+1:

Super minor, but perhaps `getTokenUrl` might be slightly more clear?
We aren't returning the actual token string here, are we?

PR #176, opened 2021-05-10
Added handling for wt_instance_error

craig-willis:
## Problem

UI does not currently handle `wt_instance_error` notification

## Approach

Bold C&P from `wt_instance_running` implementation

## How to Test

* Depends on https://github.com/whole-tale/gwvolman/pull/145 and https://github.com/whole-tale/girder_wholetale/pull/480
* Modify `setup_girder.py` to set the command for JupyterLab to something that doesn't exist (e.g., `thisdoesnotexist`)
* `make clean && make dev`
* Create  JupyterLab tale and Run
* Note progression of notifications (building image, creating volume, starting container)
* Job should fail with "Error: starting container"
* View logs, note stack trace about executable not on PATH
* Run button should change to "Error!" and "Tale launching" > "Something went wrong."


Xarthisius (approved):

bodom0015 (approved):
LGTM :+1:

Thank you for adding this.. I'll make a subsequent PR for `wt_instance_deleted`

PR #179, opened 2021-05-13
fix: view for a running tale should display interact

Xarthisius:
## Problem

View tale when tale is running should display interact page, not metadata

## How to Test

1. Run a Tale
2. Click on the view button.


bodom0015 (approved):
:+1: thank you for submitting this one. I also added a small change so that Public Tales go to Metadata (since Interact is not guaranteed to work for all Public Tales)

craig-willis (approved):

PR #187, opened 2021-05-18
feat: add progress indicator for file uploads

### LINKED ISSUE #167: File upload progress bar not displaying
**Steps to Reproduce**

* Create a new tale
* Open dev console
* Upload a large file

**Expected Results**

Upload progress bar displays during upload

**Actual Results**

Progress written to console, but no progress bar

**NOTES**

https://github.com/whole-tale/ngx-dashboard/blob/39a0b122d7a7a69ef2d18af3a7d104bfed31e636/src/app/%2Brun-tale/run-tale/tale-files/tale-files.component.ts#L229-L230

bodom0015:
## Problem
File uploads do not show their progress

Fixes #167 

## Approach
* Add a basic progress bar that shows a simple percentage (no estimated time or granular byte count, nothing fancy)

## How to Test

3. Navigate to Run Tale > Files
4. Upload a File (larger than 200MB or so)
    * You should see a progress bar and percentage now appear above the file Size as it uploads
5. Wait for the file to finish uploading
    * You should see the progress bar update over time
    * You should see the progress bar disappear once the file upload is complete

### DIFF
--- a/src/app/+run-tale/run-tale/tale-files/tale-files.component.ts
+++ b/src/app/+run-tale/run-tale/tale-files/tale-files.component.ts
@@ -25,6 +25,9 @@ import { RegisterDataDialogComponent } from '../modals/register-data-dialog/regi
 import { SelectDataDialogComponent } from '../modals/select-data-dialog/select-data-dialog.component';
 import { TaleWorkspacesDialogComponent } from '../modals/tale-workspaces-dialog/tale-workspaces-dialog.component';
 
+// import * as $ from 'jquery';
+declare var $: any;
+
 const URL = window['webkitURL'] || window.URL;  // tslint:disable-line
 
 // TODO: Is there a better place to store these constants?
@@ -190,6 +193,8 @@ export class TaleFilesComponent implements OnInit, OnChanges {
     const numChunks = Math.ceil(upload.size / chunkSize);
     this.logger.info(`Beginning upload of ${upload.name} (${upload.size} bytes in ${numChunks} chunks):`, upload);
 
+    const existing: FileElement = this.files.value.find(file => uploadId === file._id);
+
     for (let offset = 0; offset < upload.size; offset += (chunkSize + 1)) {
       // Upload the file blob one slice/chunk at a time
       const start = offset;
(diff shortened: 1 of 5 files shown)


craig-willis:
I wasn't expecting to see this one fixed, but do have a couple of minor comments.  What you've implemented is a great improvement, so these can be addressed later if at all.

I like the inline progress bar but note that it disappears if I change tabs.  Also noting that there's no way to cancel the upload (not that I could before, but it would be nice).

bodom0015:
> I like the inline progress bar but note that it disappears if I change tabs.
This is true - if you navigate away from the view, your `tale-files` component goes out of scope and your uploads are all aborted. I can (possibly) add a warning modal here to prevent the user navigating away by mistake, but in general this will be the behavior with most SPAs unless special handling is added.

This was the case before as well, even if the aborted uploads were not indicated to the user in any way

> Also noting that there's no way to cancel the upload (not that I could before, but it would be nice).
This is perhaps a question for @Xarthisius ... does the Girder API offer a way to cancel an in-progress file upload?

Xarthisius:
> This is perhaps a question for @Xarthisius ... does the Girder API offer a way to cancel an in-progress file upload?

There's `DELETE /file/upload/{id} Cancel a partially completed upload.` but I have no idea how to use it.

Xarthisius:
> I suppose, but simply stopping an incomplete upload it leaves a zero-byte file there and that is kind of ugly. This file can be deleted with the usual "Remove" button, but would be nice to not have to do this manually for each file upload that was stopped

I think that's gonna happen even if we call DELETE /file/upload/{id}. The entire flow for filesystem assetstore that we use looks roughly like this:

1. Create a file object and associate upload object with it
2. Chunks are uploaded using upload object into temporary folder.
3. Once upload is completed actual file is moved from tmp to place where it's supposed to live on fs assetstore backend.

Cancel upload AFAICT does two things: 1) removes upload object 2) removes anything associated with the upload from tmp dir. It doesn't remove the file object that was created in step 1.

bodom0015:
> Cancel upload AFAICT does two things: 1) removes upload object 2) removes anything associated with the upload from tmp dir. It doesn't remove the file object that was created in step 1.

I feel like we could call `DELETE /file/upload/{id}` and then `DELETE /file/{id}`, but I feel like these may have two distinct IDs... If they're the same or relate back to each other in some way, we could perhaps do something like the following?
```typescript
      this.fileService.fileCancelUpload(element._id).subscribe(resp => {
        this.fileService.fileDeleteFile(element._id);
      });
```

This would cancel the upload, clear out the stale `/tmp` stuff from this upload, and then delete the file created in step 1. 

Either way I think that the bigger question is, as you said, how to "stop sending chunk to the server".. RxJS has mechanisms for aborting HTTP requests, but these do not currently lend well to how the upload/chunking code is currently laid out. I can play around with being able to cancel it and/or navigate around during the upload.

That said, the current Cancel button implementation (in EmberJS) also seems a little buggy, as it does not always appear.

For example, first upload did not show a `Cancel` button:
Screen Shot 2021-05-19 at 12 37 22 PM

The next upload **did** actually show a `Cancel` button:
Screen Shot 2021-05-19 at 12 37 39 PM

craig-willis:
Thanks for looking into it. This all sounds complex enough to me that we should defer until later.

Xarthisius:
I think all the extra things we discussed above (canceling etc) should be deferred. LGTM as-is now.

Xarthisius (approved):

craig-willis (approved):

PR #188, opened 2021-05-18
fix: only prompt for update metadata if tale is shared, force cancelEdit if they confirm

### LINKED ISSUE #181: "Tale was updated" modal not behaving as expected
**Steps to Reproduce**

This could be due to my usual latency issues

* Open/create a tale
* Select Metadata page
* Open version panel and create a version
* Select "Edit" on metadata page

**Expected Results**

No "Tale was updated" modal.  Perhaps we don't display it if the tale isn't shared?

**Actual Results**

Tale was updated modal is displayed.

**NOTES**

### LINKED ISSUE #182: Changes to shared tale metadata are not updated for shared user in edit mode
**Steps to Reproduce**

1. Login as user1 and user2 in separate browser sessions
1. As user2, select "Shared with me"
1. As user1, share a tale with user2; as user2 confirm the tale appears
1. As user2, select "View" on the shared tale
1. As user1 and 2 edit the tale metadata
1. As user1 change title, author, category, description; save changes
1. As user2 confirm Tale edited modal, select Yes, confirm updates to metadata

**Expected Results**

All edited metadata fields should be updated for user2

**Actual Results**

Only the title displayed in the tale header is updated, everything else is unchanged.

bodom0015:
## Problem
The prompt to update the Tale metadata can be rather... overzealous. 

Fixes #181 
Hopefully Fixes #182 

## Approach
* Try out only showing it if the Tale has been shared with one other user or at least one group (fixes #181)
* Force `cancelEdit` if user confirms (this hopefully also fixes #182)

## How to Test
Repeat test steps reported in #181 and #182

### DIFF
--- a/src/app/+run-tale/run-tale/run-tale.component.html
+++ b/src/app/+run-tale/run-tale/run-tale.component.html
@@ -106,7 +106,7 @@
                                     </div>
 
                                     <div *ngSwitchCase="'metadata'">
-                                        <app-tale-metadata [tale]="tale" [creator]="creator"></app-tale-metadata>
+                                        <app-tale-metadata [tale]="tale" [creator]="creator" [collaborators]="collaborators"></app-tale-metadata>
                                     </div>
 
                                     <div *ngSwitchCase="'sharing'">
--- a/src/app/+run-tale/run-tale/tale-metadata/tale-metadata.component.ts
+++ b/src/app/+run-tale/run-tale/tale-metadata/tale-metadata.component.ts
@@ -23,6 +23,24 @@ import { ConfirmationModalComponent } from '@shared/common/components/confirmati
 // import * as $ from 'jquery';
 declare var $: any;
 
+// FIXME: Duplicated code - see tale-sharing.component.ts
+interface CollaboratorList {
+  users: Array<Collaborator>;
+  groups: Array<Collaborator>;
+}
+
+// FIXME: Duplicated code - see tale-sharing.component.ts
+interface Collaborator {
+  id: string;
+  level: number;
+  login: string;
+  name: string;
+  flags?: Array<any>;
+  showAlert?: boolean;
+  gravatar_baseUrl?: string;
+  affiliation?: string;
+}
+
 interface TaleAuthorValidationError {
   index: number;
   message: string;
@@ -36,6 +54,7 @@ interface TaleAuthorValidationError {
 export class TaleMetadataComponent implements OnInit {
   @Input() tale: Tale;
   @Input() creator: User;
+  @Input() collaborators: CollaboratorList;
 
   licenses: Observable<Array<License>>;
   environments: Observable<Array<Image>>;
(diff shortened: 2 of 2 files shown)


craig-willis (approved):
This works for me.  I like that the sync cancels the edit instead of trying to update the form fields. I don't see the sync modal now unexpectedly in my tests.

The only edge/hypothetical scenario that I can think of is if I try to edit the metadata prior to the completion of AinWT and the import process modifies an editable field, I could overwrite the imported info. Since `detectChanges` seems to look at all fields, maybe at some point we could limit it to look only at editable fields?

bodom0015:
Hmmmmm... While that seems like an interesting edge case, it does present a bit of a problem with this workflow... I would think that if the import process is truly that volatile, then we probably shouldn't even present the Tale to the user until it is ready.

I think the initial idea here was that a user **could** update the metadata while stuff is still importing/registering/etc. If that is not possible due to other limitations in the system, then I fail to see a reason to even present this Tale to the user yet at all when they won't be able to interact or edit it even slightly?

Maybe we could do something similar to what we did for `InstanceStatus.DELETING`, and simply hide these volatile Tales from the UI until they are stable?

craig-willis:
Agreed. It's the only case I could think of and very edgy. If it ever comes up we can revisit.

Xarthisius:
I'm following the steps from #182 and it's not canceling the edit for user2 for me. I can see that notification goes through cause tab header updates, but fields in the form stay the same and there's no  "updated tale" modal.

craig-willis:
Hmm. I thought that's how I tested it but likely didn't hard refresh for user2 incognito window.  Retesting I'm seeing the same results as @Xarthisius.  Even though user2 has edit permissions, there is no sharing tab so maybe no collaborators?

bodom0015:
Whoops, very good catch! I had forgotten to remove a bit of the logic there that was only fetching collaborators if you had Admin privileges for the Tale... Should be fixed now and hopefully works more reliably in both directions 🤞

craig-willis:
This change doesn't work for me.  I now see a bunch of 403s in the console for requests to `tale/{id}/access` for user2.  If I'm understanding, even with edit access, the user the tale is shared with doesn't have permission to read the access endpoint.

Should they? I think it's reasonable for collaborators with edit permissions to see who has access to a tale, but not change it.

Xarthisius:
> This change doesn't work for me. I now see a bunch of 403s in the console for requests to `tale/{id}/access` for user2. If I'm understanding, even with edit access, the user the tale is shared with doesn't have permission to read the access endpoint.
> 
> Should they? I think it's reasonable for collaborators with edit permissions to see who has access to a tale, but not change it.

Then we need to change: https://github.com/whole-tale/girder_wholetale/blob/2093b96a398846bea3f38384ea8be5922282c2ce/server/rest/tale.py#L397 to `DATA_READ`

Xarthisius (approved):
Works great (with whole-tale/girder_wholetale#486)!

craig-willis:
Ditto.

PR #189, opened 2021-05-18
fix: refresh tale-versions panel when recv wt_tale_updated

### LINKED ISSUE #161: Versions not refreshed when tale updated
**Problem**:
This is related to https://github.com/whole-tale/ngx-dashboard/issues/121 and will likely be considered an edge case.
When user1 creates a version on a shared tale, the new version is not reflected for user2 until after refresh.

**Test**:
* Create a tale as user1 and share with user2
* As user1 and user1 access the version panel
* User1 creates a new version

**Expected**:
User2's Version panel for is updated to reflect the new version. This should also happen for renames and deletes.

**Actual**:
`wt_tale_updated` event is sent, but version panel is not updated.

**Workaround**:
Refresh or close/reopen version panel.

bodom0015:
## Problem
Tale Versions panel is not refreshed when we receive a `wt_tale_updated`, even though it is sent when a version is created.

Fixes #161 

## Approach
* Add special handling in the Tale Versions panel to wire up to the SyncService and refresh the version list when we receive a `wt_tale_updated` event

## How to Test

3. Navigate to the Run Tale view
4. Expand the Versions panel
5. Open a fresh incognito window and login and view the same Tale
5. Save a New Version on one of the two open tabs
6. Switch to the other tab
    * You should see the new version appear that was created outside of this tab

### DIFF
--- a/src/app/+run-tale/run-tale/tale-versions-panel/tale-versions-panel.component.ts
+++ b/src/app/+run-tale/run-tale/tale-versions-panel/tale-versions-panel.component.ts
@@ -12,7 +12,8 @@ import { WindowService } from '@framework/core/window.service';
 import { enterZone } from '@framework/ngrx/enter-zone.operator';
 import { NotificationService } from '@shared/error-handler/services/notification.service';
 import { TaleAuthor } from '@tales/models/tale-author';
-import { Observable } from 'rxjs';
+import { SyncService } from '@tales/sync.service';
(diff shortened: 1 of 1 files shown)


craig-willis (approved):
WFM also with test case described in https://github.com/whole-tale/ngx-dashboard/issues/161

Xarthisius (approved):

PR #190, opened 2021-05-18
feat: indicate tale import status via the Run button

### LINKED ISSUE #121: Handling stale/changed data in UI
The goal of this issue is to start a discussion about 1) how we can address this and 2) identify events that should be handled.

**Problem**: With the move to Angular, we no longer have an Ember-data-like feature to handle updating data in the UI that has changed by other means.  This was discussed in the context of sharing (https://github.com/whole-tale/ngx-dashboard/pull/16#issuecomment-729154698), but is also apparent in other areas as well.  As I recall from a dev meeting, one way to handle this is to use the notification stream to inform the UI of changes. Currently, it often requires a complete refresh of the page:

I am aware of the following:
* When a tale is shared with me, it does not appear on the Tales page. This is probably the case when a tale is made public by another user
* When another user makes changes to workspace, external data,  tale metadata, creates a version, etc. these changes are not reflected in my UI without refresh.
* When I import a tale (e.g., from Zenodo), the initial tale is created but the resulting imported data is not reflected without refresh
* ...?

### LINKED ISSUE #137: Handle Tale state changes during in-progress jobs
**Steps to Reproduce**

1. Import a tale from Zenodo (e.g., Click "Run this on Whole Tale" for https://sandbox.zenodo.org/record/718277)
2. Note that a Tale is created and you are redirected to the metadata page, but metadata is missing ~involatile data~ datasets used/related identifiers
3. A job notification appears for the import process (import notifications require https://github.com/whole-tale/girder_wholetale/pull/445)
4. Inspect Tale status during import process (via Swagger or otherwise) and value should be PREPARING (0) until the job completes (READY).
5. When import is complete, the tale metadata page is unchanged
6. Refresh, note that Involatile Data section is updated

**Expected Results**
* The UI indicates that the Tale is in a pending state
* When the Tale state changes to READY (1) the metadata page is refreshed/updated

**Actual Results**
* Tale state changes are not reflected in UI and incomplete metadata is displayed

bodom0015:
## Problem
Run View contains stale data after Tale import completes - needs manual refresh to pick up newly-registered datasets, related identifiers, etc

Fixes #137 
Fixes #121 

## Approach
* Wire up the Run Tale view to the SyncService to refresh Tale data when import completes for the Tale

## How to Test
3. Import a Tale (preferably a large Tale that take awhile to import and/or ultimately fails)
    * You should see that the "Run Tale" button is disabled and reads `Importing...`
4. Wait for the Tale import to finish
    * On failure, you should see that the button stays disabled and reads `Import Failed`
    * On success, you should see the "Run Tale" button now appears as expected, allowing the user to interact with it to launch the Tale

### DIFF
--- a/src/app/tales/components/tale-run-button/tale-run-button.component.html
+++ b/src/app/tales/components/tale-run-button/tale-run-button.component.html
@@ -1,22 +1,32 @@
 <div>
-  <button *ngIf="!instance" class="right floated" [ngClass]="{ 'tale-button': !isPrimary, 'tale-button-primary': isPrimary }" (click)="startTale()">
-      <i class="play icon"></i>
-      Run Tale
-  </button>
-  <button *ngIf="instance && instance.status === 0" class="right floated" [ngClass]="{ 'tale-button': !isPrimary, 'tale-button-primary': isPrimary }" (click)="startTale()">
+  <button *ngIf="tale && tale.status === 1" class="right floated" [ngClass]="{ 'tale-button disabled': !isPrimary, 'tale-button-primary disabled': isPrimary }" disabled="true">
       <i class="fas fa-spinner fa-pulse"></i>
-      Starting...
+      Importing...
   </button>
-  <button *ngIf="instance && instance.status === 1" class="right floated" [ngClass]="{ 'tale-button': !isPrimary, 'tale-button-primary': isPrimary }" (click)="stopTale()">
-      <i class="stop icon"></i>
-      Stop Tale
-  </button>
-  <button *ngIf="instance && instance.status === 2" class="right floated" [ngClass]="{ 'tale-button': !isPrimary, 'tale-button-primary': isPrimary }" (click)="stopTale()">
+  <button *ngIf="tale && tale.status === 2" class="right floated" [ngClass]="{ 'tale-button disabled': !isPrimary, 'tale-button-primary disabled': isPrimary }" disabled="true">
       <i class="warning icon"></i>
-      Error
+      Import Failed
   </button>
-  <button *ngIf="instance && instance.status === 3" class="right floated" [ngClass]="{ 'tale-button': !isPrimary, 'tale-button-primary': isPrimary }" (click)="stopTale()">
-      <i class="fas fa-spinner fa-pulse"></i>
-      Stopping...
-  </button>
+  <div *ngIf="tale && tale.status === 2">
+      <button *ngIf="!instance" class="right floated" [ngClass]="{ 'tale-button': !isPrimary, 'tale-button-primary': isPrimary }" (click)="startTale()">
+          <i class="play icon"></i>
+          Run Tale
+      </button>
+      <button *ngIf="instance && instance.status === 0" class="right floated" [ngClass]="{ 'tale-button': !isPrimary, 'tale-button-primary': isPrimary }" (click)="startTale()">
+          <i class="fas fa-spinner fa-pulse"></i>
+          Starting...
+      </button>
+      <button *ngIf="instance && instance.status === 1" class="right floated" [ngClass]="{ 'tale-button': !isPrimary, 'tale-button-primary': isPrimary }" (click)="stopTale()">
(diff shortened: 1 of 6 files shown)


craig-willis:
I think the stale "datasets used" and "related identifiers" were already handled in an earlier PR with the `wt_tale_updated` implementation, although I see no harm in specifically handling the completed event. For me what remains of https://github.com/whole-tale/ngx-dashboard/issues/137 is some indication that the tale import is incomplete.

bodom0015:
Ah, sorry.. I think I had misread your comment here and thought there was an edge case here that we weren't handling: https://github.com/whole-tale/ngx-dashboard/issues/137#issuecomment-799668756

bodom0015:
Updated this PR to show the Tale Import status on the Run view, and updated test case above.

After successful import, you should see the "Run Tale" button appear as it does currently

During import, you should see a disabled button with a spinner that reads `Importing...`:
Screen Shot 2021-05-25 at 3 37 42 PM

On a failed import, you should see the button is still disabled (open question: how does the user resolve a failed import?):
Screen Shot 2021-05-25 at 3 38 07 PM

craig-willis (approved):
I tested using https://dashboard.local.wholetale.org/browse?uri=https%3A%2F%2Fdoi.org%2F10.5281%2Fzenodo.4774875&name=Test and selecting READ/WRITE, the copy workspace takes a little time.  I previously used https://doi.org/10.5281/zenodo.4782234, which is very large and caused an error during import.  I see the button state changes as expected.  

> how is the user expected resolve a failed import? 

My expectation is that they delete and reimport (for now)

My initial reaction was that "Importing..." and "Import failed" are too specific, given other processes may change the tale status.  The only other call that I see is `copyTale`, but it doesn't have notifications, so maybe not worrying about for now.

Part of me thinks "Importing..." and "Import failed" are too specific. Looking at girder_wholetale, most of the tale status changes are from import tasks, but also copyTale. But copy doesn't have notifications

bodom0015:
The labels on the Girder side are `PREPARING` instead of "Importing...", and `ERROR` instead of "Import Failed"

I am unclear on when else these labels would be applied, outside of the Import. For example, currently if the build fails the user can still click the Run button to retry the build - the same is not true with Import, which appears to be the outlying special case here?

I'm certainly open to changing the wording there if there are cases outside of import when this would be applicable

src/app/tales/components/tale-run-button/tale-run-button.component.html
```diff
@@ -1,22 +1,32 @@
...
\ No newline at end of file
+<div>
+  <button *ngIf="tale && tale.status === 0" class="right floated" [ngClass]="{ 'tale-button disabled': !isPrimary, 'tale-button-primary disabled': isPrimary }" disabled="true">
+      <i class="fas fa-spinner fa-pulse"></i>
+      Importing...
+  </button>
+  <button *ngIf="tale && tale.status === 2" class="right floated" [ngClass]="{ 'tale-button disabled': !isPrimary, 'tale-button-primary disabled': isPrimary }" disabled="true">
+      <i class="warning icon"></i>
+      Import Failed
+  </button>
```
bodom0015, line 9:
```suggestion
  <button *ngIf="tale && tale.status === 0" class="right floated" [ngClass]="{ 'tale-button disabled': !isPrimary, 'tale-button-primary disabled': isPrimary }" disabled="true">
      <i class="fas fa-spinner fa-pulse"></i>
      Preparing...
  </button>
  <button *ngIf="tale && tale.status === 2" class="right floated" [ngClass]="{ 'tale-button disabled': !isPrimary, 'tale-button-primary disabled': isPrimary }" disabled="true">
      <i class="warning icon"></i>
      Initialization Failed
  </button>

```
Note that the "Error" label is already applied in some cases where the instance fails to launch. Such cases are correctable by deleting and relaunching (e.g. pressing the Error button and then clicking Run Tale again).

To distinguish such events from failed imports / failed copies, I'd suggest using something slightly more explicit here

PR #192, opened 2021-05-24
fix: don't access collaborators list as ro user

### LINKED ISSUE #191: Accessing Tale collaborators list as read-only users fails with 403.
per https://github.com/whole-tale/wt-design-docs/issues/155:
When shared user has view only permissions, lots of console 403s on calls to access endpoint along with big trace. Not certain, but this error may affect UI behavior

**Steps to Reproduce**

1. During deployment create extra users via `python scripts/create_extra_users.py`
2. Log as yourself, create a Tale, share it "view only" with extra users
3. In a separate tab as an extra user navigate to "shared with me", click view on a Tale.

**Expected Results**

No angry tracebacks in the dev console.

**Actual Results**

```
 main.c7f0d694c1c23437436d.js:1 ERROR Error: Uncaught (in promise): e: {"headers":{"normalizedNames":{},"lazyUpdate":null},"status":403,"statusText":"OK","url":"https://girder.stage.wholetale.org/api/v1/tale/60a866713a4e9f14ff07dcba/access","ok":false,"name":"HttpErrorResponse","message":"Http failure response for https://girder.stage.wholetale.org/api/v1/tale/60a866713a4e9f14ff07dcba/access: 403 OK","error":{"message":"Write access denied for tale 60a866713a4e9f14ff07dcba (user 59fb6165f7e8790001da4e8b).","type":"access"}}
```

**NOTES**

`GET /tale/:id/access` requires at least `WRITE` access level for user to get the list of collaborators.

Xarthisius:
Fixes #191

## Approach
Don't do things that are forbidden

## How to Test

See #191

### DIFF
--- a/src/app/+run-tale/run-tale/run-tale.component.ts
+++ b/src/app/+run-tale/run-tale/run-tale.component.ts
@@ -138,6 +138,9 @@ export class RunTaleComponent extends BaseComponent implements OnInit, OnChanges
     }
 
     refreshCollaborators(): Promise<any> {
+      if (!this.tale || this.tale._accessLevel < AccessLevel.Write) {
+        return new Promise(resolve => {});
+      }
       return this.taleService.taleGetTaleAccess(this.taleId).toPromise().then((resp: any) => {
         this.logger.info("Fetched collaborators:", resp);
         this.collaborators = resp;


craig-willis (approved):
That fixed it for me.

bodom0015 (approved):
LGTM :+1:

Good catch! Sorry if this was introduced by a hasty that I should've thought about and tested more thoroughly.

Issue #193, opened 2021-05-24
'api' parameter not handled by dashboard


craig-willis:
**Steps to Reproduce**

From test plan, import the following datasets from non-production DataONE coordinating nodes (CN):

* "proveit": https://dashboard.stage.wholetale.org/browse?api=https%3A%2F%2Fcn-stage-2.test.dataone.org%2Fcn%2Fv2&name=proveit&uri=https://dev.nceas.ucsb.edu/view/doi:10.5072/FK2K075M25
* "Alpha Helix 2001...": https://dashboard.stage.wholetale.org/browse?api=http%3A%2F%2Fcn-stage.test.dataone.org%2Fcn%2Fv2&name=Alpha%20Helix&uri=https://test.arcticdata.io/view/doi:10.5065/D6N014NC

Note that the integration API redirects from 

https://girder.stage.wholetale.org/api/v1/integration/dataone?api=https%3A%2F%2Fcn-stage-2.test.dataone.org%2Fcn%2Fv2&title=proveit&uri=https://dev.nceas.ucsb.edu/view/doi:10.5072/FK2K075M25 

to 

https://dashboard.stage.wholetale.org/browse?uri=https%3A%2F%2Fdev.nceas.ucsb.edu%2Fview%2Fdoi%3A10.5072%2FFK2K075M25&name=proveit&api=https%3A%2F%2Fcn-stage-2.test.dataone.org%2Fcn%2Fv2

**Expected Results**

Tale created and dataset is imported and available under external data

**Actual Results**

* Tale is created, but datasets are incorrectly imported as HTTP resources
* API parameter is ignored

**Notes**
* I expect that we want any integrators (e.g., DataONE sites) to use the `v1/api/integration/dataone` call and not go direct to browse.
* `/browse` route no longer exists and part of its legacy is likely this integration. We could fix that too.

PR #201, opened 2021-06-01
Allow to import Tales/Binders from testing DataONE deployments

Xarthisius:
## Problem

See #193 

## Approach
Parse `api` query parameter and send it back as `lookupKwargs.base_url`

## How to Test

Explained in #193 (change `.stage` to `.local` in all urls)


craig-willis (approved):
Working for me.

bodom0015 (approved):
LGTM :+1:

PR #207, opened 2021-06-23
fix: disable "Restart Tale" when Tale is stopped

### LINKED ISSUE #199: User shouldn't be able to restart stopped Tale
**Steps to reproduce**

1. Open a stopped tale
2. Select tale "..." menu

**Expected results**
Restart option is disabled/not present

**Actual results**
Restart option is present and does nothing (no error)

bodom0015:
## Problem
"Restart Tale" is offered even when Tale is not running

Fixes #199 

## Approach
* Hide this option from the dropdown
* Also make the icon spacing consistent on this dropdown (it was bothering me)

NOTE: I tried simply disabling this option, but that doesn't allow me to show a tooltip since it is `disabled` ... so there was no way how to let the user know why the item was disabled.

## How to Test
Prerequisites: at least one Tale created, not running

3. Navigate to Run Tale for a stopped Tale
4. Expand the dropdown at the top-right
    * You should **not** see the "Restart Tale" option listed here
5. Run the Tale, wait for Tale to fully start (e.g. `instance.status === 1`)
6. Expand the dropdown again
    * You should see the "Restart Tale" option is now listed here


src/app/+run-tale/run-tale/run-tale.component.html
```diff
@@ -1,130 +1,133 @@
...
+                                        Close
+                                    </a>
+
+                                    <div id="tale-actions-dropdown"  class="ui floating tale dropdown theme" tabindex="0">
+
+                                        <mat-icon fontSet="fa" fontIcon="fa-ellipsis-v"></mat-icon>
+
+                                        <div class="menu transition hidden" tabindex="-1">
+                                            <div class="item" (click)="rebuildTale()" *ngIf="isTaleWritable()"><i class="fas fa-fw fa-redo-alt"></i> Rebuild Tale</div>
+                                            <div class="item" (click)="restartTale()" *ngIf="isTaleWritable() && instance && instance.status === -1">
```
craig-willis, line 49:
```suggestion
                                            <div class="item" (click)="restartTale()" *ngIf="isTaleWritable() && instance && instance.status === 1">

```

Shouldn't this be 1?

craig-willis (changes requested):

src/app/+run-tale/run-tale/run-tale.component.html
bodom0015: Good catch! Indeed, this should probably be a 1

This is why I shouldn't make PRs at 3AM :X lol

Xarthisius (approved):

craig-willis (approved):

PR #213, opened 2021-07-12
Fix blank page error in Docker

bodom0015:
## Problem
I'm not sure how this wasn't caught during my own testing, but the Docker container seems to only bring up a blank page.

This was likely a result of the Angular 9 upgrade, as lots of things moved around there.

## Approach
* Set default `LogLevel` to `Error` - previous value of `Info` was somehow hiding valuable error messages in the console
* Move `ApiModule` import from `AppModule` to `SharedModule`, fetch `env.json` from static file when initializing `WindowService` - this fixed a circular dependency
* Replace `env.js` with a static JSON file `env.json` - this seemed to behave differently when using `ng build --prod` vs `ng build`?

## How to Test
Setup:
Because the `entrypoint.sh` has changed, we need to build a new image:
1. `docker build -t yourdockeruser/ngx-dashboard:post-migration-fix .`
2. `docker push yourdockeruser/ngx-dashboard:post-migration-fix`
3. `docker service update wt_dashboard --image yourdockeruser/ngx-dashboard:post-migration-fix --force`

Test Steps:
    * You should be able to login and browse the platform as usual

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -54,4 +54,4 @@ package-lock.json
 Thumbs.db
 
 
-src/assets/env.js
+src/assets/env.json
--- a/angular.json
+++ b/angular.json
@@ -50,8 +50,8 @@
               "src/assets/data",
               "src/assets/i18n",
               "src/assets/config.local.json",
-              "src/assets/env.template.js",
-              "src/assets/env.js"
+              "src/assets/env.template.json",
+              "src/assets/env.json"
             ],
             "styles": [
               "src/assets/sass/lib.scss",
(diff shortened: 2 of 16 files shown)


craig-willis:
I'm able to login and browse. I haven't tested exhaustively, but am seeing other errors:
* Selecting "Create New Tale" results in lots of "Can't bind to 'mat-dialog-close' since it isn't a known property of 'button'". I can't create or cancel the tale. I get the same error on delete tale, which makes me think this happens in other places too.
* Uploading a file into the file browser results in "local-console.service.ts:33 Error: unable to get icon for FileElement - unrecognized modelType:". The file uploads, but for me the context menu no longer works.

bodom0015:
Good catch! I pushed one fix for this yesterday, but my internet dropped at 5pm before I could push the second one

Both have been pushed up now :+1:

Xarthisius (approved):

craig-willis (approved):
lgtm.

PR #217, opened 2021-08-09
Instance running should update interact tab

bodom0015:
## Problem
When a Tale finishes starting up when the user is on the Interact tab, the view does not refresh automatically to show the Tale.

Instead, the view should automatically update (and not require navigation/refresh) to display the running Tale.

## Approach
The `tale-run-button` lives on the Run View (as well as on the Catalog View). This component takes care of starting / stopping Tales and previously held the logic for polling for instance status.

I think the issue here was that when an update was received from the SyncService, `emit()` was not being called to output this status change to the parent component.

Now, `tale-run-button` should call `.emit()` to `@Output` the new instance status to the `run-tale` parent component as expected.

I also improved some of the logging to print out the actual status string when updating the instance status

## How to Test
Prerequisites: at least one Tale created by you

3. Start a Tale
    * You should automatically be brought to the Run > Interact view
    * You should see a loading spinner indicating that the Tale is `Launching` 
4. Wait for the Tale to finish launching
    * You should see the loading spinner disappear 
    * You should see the running Tale appear in the view

### DIFF
--- a/src/app/tales/components/tale-run-button/tale-run-button.component.ts
+++ b/src/app/tales/components/tale-run-button/tale-run-button.component.ts
@@ -12,7 +12,7 @@ import { Subscription } from 'rxjs';
 @Component({
   selector: 'app-tale-run-button',
   templateUrl: './tale-run-button.component.html',
-  styleUrls: ['./tale-run-button.component.scss']
+  styleUrls: ['./tale-run-button.component.scss'],
 })
 export class TaleRunButtonComponent implements OnInit, OnChanges, OnDestroy {
   @Input() instance: Instance;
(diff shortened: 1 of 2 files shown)


src/app/tales/components/tale-run-button/tale-run-button.component.ts
```diff
@@ -65,6 +81,7 @@ export class TaleRunButtonComponent implements OnInit, OnChanges, OnDestroy {
     this.instanceService.instanceGetInstance(resource.instanceId).subscribe((instance: Instance) => {
       this.instance = instance;
       this.ref.detectChanges();
+      this.taleInstanceStateChanged.emit(this);
```
bodom0015, line 84: This was the bit that was missing

Xarthisius (approved):

craig-willis (approved):
Sorry, thought I already approved this.

Issue #230, opened 2021-10-26
Add admin option for tale sharing


craig-willis:
With https://github.com/whole-tale/wt_versioning/pull/36, it will no longer be possible for someone with WRITE permission to delete a version they've created.  We should add an "admin" or "owner" option for sharing that allows a user to share a tale with full permissions.

PR #231, opened 2021-10-26
Own acl

Xarthisius:
## Problem

ADMIN / Own ACL is missing. See #230 

## Approach

Adds a missing ACL that's supported by the backend, where it's necessary

## How to Test

1. Create additional users (`scripts/create_extra_users.py`)
1. Create a Tale. Navigate to Tale > Share
2. Go wild with new "Owns" ACL
3. Inspect the Tale object to confirm user has access level "2"


bodom0015 (approved):
LGTM :+1:

Also just noticed when testing this that when no gravatar is set for the user, a broken image is shown. Filed this as #232 and will address this with a small PR to correct this (to just show the WT logo) in such cases.

PR #243, opened 2022-01-26
Tale Version Panel fixes: orderby created date, link directly to runs/versions

### LINKED ISSUE #234: Tale Versioning improvements
Found during review of #205

* Renaming a version changes order in history panel (should be based on created date instead?)
* Renamed version is not reflected in "Run based on "
* Link to version from "Run based on..." is nice, can we do the same for the version and run names (i.e., the names link to the version/run in the file browser)?

bodom0015:
## Problem
Fixes #234 

## Approach
* Order by `created` date, instead of `updated` date
* ??? - fix for the rename issue may need backend signal work, or else the `SyncService` is missing some implementation
* Link to Run > Files > Tale Versions and/or Run > Files > Recorded Runs directly from the Tale History Panel

## How to Test
Prerequisites: at least on Tale created, at least one Tale Version created, at least one Run created from the Version

3. Navigate to Run Tale and create a Version and a Run (if you haven't already
4. Expand the Tale History Panel and view the timeline
   * You should see timeline is now sorted by `created` date
   * You should see that the version/run name is now a link
5. Rename the version
    * You should see the order does not change in the list (e.g. the run should be created after version, no matter what) 
6. Click on the version name
    * You should be brought to Run > Files > Tale Versions with the chosen version highlighted
6. Click on the run name
    * You should be brought to Run > Files > Recorded Runs with the chosen run highlighted

### DIFF
--- a/src/app/+run-tale/run-tale/tale-versions-panel/tale-versions-panel.component.html
+++ b/src/app/+run-tale/run-tale/tale-versions-panel/tale-versions-panel.component.html
@@ -51,14 +51,19 @@
           </div>
         </div>
         <div class="content">
-          <div class="date">{{ res.updated | date:'short' }}</div>
+          <div class="date">{{ res.created | date:'short' }}</div>
           <div class="header">
             <img class="ui avatar image" *ngIf="(res | taleCreator | async) as creator" [src]="creator.gravatar_baseUrl" title="{{ creator.firstName }} {{ creator.lastName }} ({{ creator.login }})" />
+
             <span *ngIf="!res.runVersionId">Version saved:</span>
             <span *ngIf="res.runVersionId">Recorded run:</span>
           </div>
 
-          {{ res.name }}
+          <a href="#" [routerLink]="'/run/' + tale._id"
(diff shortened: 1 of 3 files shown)


craig-willis:
Linking to versions/runs looks good and works for me, but not ordering.  

Here's what I see:
* Create version with default (timestamp) name
* Create another version with default (timestamp) name
* Versions should be ordered with newest first in history panel
* Rename first version to "test"
* This version now appears at the top of the panel

This happens with both versions and runs.

bodom0015:
Thanks for the feedback, folks! ^_^

I actually found the sort function this time... sorry for the very silly oversight there, but renaming things no longer changes the order of the timeline :+1:

I also reworked the `TaleVersionNamePipe` to be more observant and more performant. Passing in a reference of the timeline to the pipe causes the pipe to re-render the version name when the timeline itself changes. However, this actually presents a performance benefit as well: by passing in the `timeline` from the parent component, we can simply search the timeline for the version in question and no longer need an individual fetch anymore :+1:

craig-willis (approved):
Order, links to versions/runs, and updating the history panel after renaming all work for me now.

Xarthisius (approved):

PR #252, opened 2022-02-21
Fix AiWT URL encoding / roll-back when rename fails

### LINKED ISSUE #249: Handle error when rename fails due to existing target
Items appear to be renamed in UI after error due to existing target.

See https://github.com/whole-tale/virtual_resources/pull/19

To repeat:

* Create folder1, folder2
* Rename folder2 to folder1
* Error window appears indicating folder1 exists, but name is updated in UI

### LINKED ISSUE #250: AinWT URL encoding issue when user isn't logged in
**Steps to Reproduce**

* Open a new private browser session
* https://dashboard.local.wholetale.org/browse?uri=https%3A%2F%2Fdoi.org%2F10.5281%2Fzenodo.820575&name=Automotive%20Sensor%20Data

**Expected Results**
Title and data source are not URL encoded

**Actual Results**
Title and data source are URL encoded

Screen Shot 2022-02-17 at 4 42 59 PM

**NOTES**

Go to the same URL when logged in and things work as expected

bodom0015:
## Problem
Fixes #249 - If renaming a file or folder, the UI shows the new name as if it were successful
Fixes #250 - If a user is not logged in and attempts to trigger AiWT, their URL parameters are double-encoded and are not properly decoded

## Approach
* #249: store `prevName` when renaming, roll-back to this name if the rename fails
* #250: decode all params for AiWT (in case they are double-encoded by the login redirect)

### Testing #249

2. Navigate to Run > Files (Home or Workspace) for a Tale that you own
3. Create a folder named `folder1`
4. Create a second folder named `folder`
5. Attempt to rename `folder2` to `folder1`
    * You should see that an error modal appears on the screen indicating that the rename has failed
    * You should see the previous name `folder2` is restored
    * You should **NOT** see two listings named `folder1`

### Testing #250
2. Close all incognito windows, then open a fresh incognito window
3. Navigate to https://dashboard.local.wholetale.org/browse?uri=https%3A%2F%2Fdoi.org%2F10.5281%2Fzenodo.820575&name=Automotive%20Sensor%20Data
    * You should be redirected to login
    * Notice the URL parameters are double-encoded
4. Click "Access WholeTale" and log in
    * After login, you should be redirected to the "My Tales" part of the Tale Catalog
    * You should see the Create Tale Modal automatically open
    * You should see that the `Name` field is pre-populated and that the value is not URL encoded
5. Look toward the bottom of the modal
    * You should see that the `Data Source` is populated with a URL that is properly encoded (e.g. only suffix is encoded, URL part should be readable)

### DIFF
--- a/src/app/+run-tale/run-tale/tale-files/tale-files.component.ts
+++ b/src/app/+run-tale/run-tale/tale-files/tale-files.component.ts
@@ -744,6 +744,9 @@ export class TaleFilesComponent implements OnInit, OnChanges {
       this.runService.runPutRenameRun(params.id, params.name).subscribe(resp => {
         this.load();
       }, err => {
+        // Rename failed, roll-back the change
+        element.name = element.prevName;
+        this.ref.detectChanges();
         this.dialog.open(ErrorModalComponent, { data: { error: err.error } });
       });
 
@@ -757,6 +760,9 @@ export class TaleFilesComponent implements OnInit, OnChanges {
       this.versionService.versionPutRenameVersion(params.id, params.name).subscribe(resp => {
         this.load();
       }, err => {
+        // Rename failed, roll-back the change
+        element.name = element.prevName;
+        this.ref.detectChanges();
         this.dialog.open(ErrorModalComponent, { data: { error: err.error } });
       });
 
@@ -774,6 +780,9 @@ export class TaleFilesComponent implements OnInit, OnChanges {
         folders[index] = resp;
         this.folders.next(folders);
       }, err => {
+        // Rename failed, roll-back the change
+        element.name = element.prevName;
+        this.ref.detectChanges();
         this.dialog.open(ErrorModalComponent, { data: { error: err.error } });
       });
     } else if (element._modelType === 'item') {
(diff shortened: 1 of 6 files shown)


bodom0015:
Testing things and it looks like the rollback on rename failure is not behaving as intended because the case presented does not return a failure from the API.

**Steps to reproduce**
1. Create `folder1`
2. Create `folder2`
3. Rename `folder2` to be called `folder1`

**Expected Results**
Rename operation should fail, name should be rolled back

**Actual Results**
Rename returns 200, but original `folder1` is overwritten by new `folder1`, previous folder is inaccessible... what happens to the folder contents? Does this leave stale data in Girder that is not longer contained in any folder?

Xarthisius:
> **Steps to reproduce**
> 
> 1. Create `folder1`
> 2. Create `folder2`
> 3. Rename `folder2` to be called `folder1`

I cannot reproduce that. With this PR I get a nice message "A folder or file with that name already exists here." and name is not changed. Checked both dashboard and girder.

Xarthisius (approved):

PR #257, opened 2022-02-28
Fix setting GIRDER_API_URL envvar at runtime

### LINKED ISSUE #255: Dashboard is still using hardcoded girder.local.wholetale.org
I've deployed latest to the new test cluster at https://dashboard.test.wholetale.org. The Dashboard is failing to load because it is using the API at https://girder.local.wholetale.org/api/v1/. This should've been fixed by https://github.com/whole-tale/ngx-dashboard/issues/237. 
 
```
ubuntu@wt-test-nfs:~$ docker ps | grep dashboard
0c7d4779dedc        wholetale/ngx-dashboard:latest   "/entrypoint.sh"       29 minutes ago     
       Up 29 minutes       80/tcp              wt_dashboard.1.ro6nla0wsbb66aw7wz8pd92uv
ubuntu@wt-test-nfs:~$ docker exec -it 0c7d4779dedc sh
$ echo $GIRDER_API_URL
https://girder.test.wholetale.org/api/v1
```

**Steps to Reproduce**
* Open dev console on network tab 
* Go to https://dashboard.test.wholetale.org
* Note failed requests to https://girder.local.wholetale.org/api/v1/user/me 

**Expected Results**

Dashboard loads with requests to https://girder.test.wholetale.org

**NOTES**
* I re-tested based on https://github.com/whole-tale/ngx-dashboard/pull/242 which was merged to master. The `apiUrl` value in my `env.json` is `https://www.google.com` and dashboard loads using girder.local. after empty cache/reload.
* https://github.com/whole-tale/ngx-dashboard/blob/bd306102afed088abdc85970fbe469cddfb2e188/src/app/shared/shared.module.ts#L21-L24
* main.XXX.js in-container contains `$.forRoot({rootUrl:"https://girder.local.wholetale.org/api/v1"})`

bodom0015:
## Problem
`GIRDER_API_URL` needs to be set before everything loads, but currently uses an async fetch.

This is problematic because the UI code needs this value when it first loads - setting it after that can lead to failed requests or stale/bad data in the browser.

Fixes #255 

## Approach
Our previous iteration of the dashboard did not have this problem, since it loaded the envs as raw JS instead of a JSON fetch.

Some adjustments to our templates / docker patterns made this work:
* `env.template.js` now produces `env.js` - this JS file can be loaded directly in `index.html`
* Load `env.js` in the `index.html` - this ensures that envs have been loaded before first render or any requests are sent
* Adjust entrypoint / source to use `env.js` instead of `env.json` throughout

## How to Test
2. Set `GIRDER_API_URL="https://google.com"` and start up the dashboard
3. Navigate to the WholeTale Dashboard
    * You should see failures in the console trying to access `https://google.com/users/me`
4. Shut it down and restart with `GIRDER_API_URL="https://girder.local.wholetale.org/api/v1"`
    * You should see the UI now loads as expected


craig-willis (approved):
LGTM. Slightly different test case:
* `docker build -t craigwillis/ngx-dashboard:api-fix .`
* `docker service update wt_dashboard --image craigwillis/ngx-dashboard:api-fix`
* `docker service update wt_dashboard --env-add GIRDER_API_URL=https://www.google.com`
  * Empty cache and hard reload -- note errors 
* `docker service update wt_dashboard --env-add GIRDER_API_URL=https://girder.local.wholetale.org/api/v1`
  * Empty cache and hard reload -- no errors!

craig-willis:
For me, the TOS link on the landing page points to https://dashboard.local.wholetale.org/undefined/tos.  I saw this before this fix -- but could a fix be included?

bodom0015:
@craig-willis good catch on this! Apparently, everything about `WindowService` no longer works. if you recall, we recently saw issues with calling things like `this.windowService.open('https://somewebsite.com', '_blank');` to open a new tab - that is likely when this broke as well.

I have moved ahead and deprecated WindowService, since nothing about it works as intended anymore. Using `window` directly has so far worked everywhere, and since we're just reading from it data and opening new tabs, it should not need direct Angular integration. I've also added a comment there that `window` should now be used instead of `WindowService` :+1:

I also fixed the build warnings 🎊 mostly very minor changes to the ordering of a few import statements, but it makes me feel better to clean these up before a new release

bodom0015:
Alright, scratch that odd pattern I noted.. I've added the following to our root AppModule and that seems to let the compiler/IDE and the linting rules work together in harmony:
```typescript
declare global {
  interface Window {
    readonly navigator: Navigator;
    location: Location;
    env: {
      apiUrl: string;
      dataONEBaseUrl: string;
      authProvider: string;
      rtdBaseUrl: string;
    };

    alert(msg: string): void;
    open(url: string, target?: string): void;
    confirm(msg: string): void;
  }
}
```

This merges the existing Window definition with our own to explicitly add the `env` so that the compiler knows about it . And with this I'm convinced that we no longer need the old `Window` model or `WindowService`, so I've removed those now for good as well :+1:

craig-willis (approved):
Great, that works better. TOS issue is fixed too. I ran a basic UI test (create tale, create folder/upload, run tale, edit metadata, etc) and things are working.

Issue #260, opened 2022-03-01
Disable export for users without write permission


craig-willis:
**Problem:**
Because we create a version if not present on export, users should not be allowed to export tales they don't have write access to.  They will need to copy the tale to export.  I propose disabling the export menu as we already do for publish.

**To test**:
* Create public LIGO tale  (`register_ligo.py`)
* View tale and select "..." > Export

**Expected results:**
Export menu is either not available or user is prompted to copy.

**Actual results:**
Export fails with error
```
{
    "message": "Write access denied for tale 6220ddfbf20a53f197bd1b27 (user 621d3491407ba1e0d9fecd0f).",
    "type": "access"
}
```

PR #262, opened 2022-03-03
Disable export menu option if not writable

craig-willis:
See https://github.com/whole-tale/ngx-dashboard/issues/260 for test case

This PR simply disables the Export menu as we already do publish. Another option would be to prompt the user to copy the tale.


bodom0015:
Ahhh, I did not realize that write access was needed for export, since it is a read-only operation on the server.
I agree with this change given the current behavior of the backend :+1:

That said, I'm curious - is there a reason why write access is required for the backend export? They're technically not writing to the Tale itself, right?

craig-willis:
An exported or published tale must have a version. When versioning was added, the export and publish operations create a version if not present, so the user currently needs write access. Possibly worth further discussion for a future release.

Xarthisius (approved):

bodom0015 (approved):
LGTM :+1: did not know about creating a version during export, but that makes sense

PR #265, opened 2022-03-07
Ensure correct file manager nav contents

### LINKED ISSUE #261: File manager sometimes shows content for wrong tab
**Steps to Reproduce**
I don't yet have a reproducible case for this, but have seen it often on my dev env and now on v1.1rc1 testing.  Switching tabs between workspace/external data/versions/runs, I not-infrequently see the content for the wrong tab -- forcing a refresh resolves it.  For example, my versions are showing in External Data below:

image

**Expected Results**
The right content always shows for the selected tab.

**Actual Results**

The wrong content sometimes shows for the selected tab.

bodom0015:
## Problem
The UI can sometimes show incorrect content after loading. My current thinking is that request timing is a problem here, and we are possibly fulfilling requests out of order due to their timing.

This also attempts to fix avalanching requests when switching tabs on Run Tale (@Xarthisius reproduced this during the 3/7 dev meeting)

Fixes #261 

## Approach
* Add conditionals to the Run > Files view - ignore requests that aren't for our current nav when the request finishes
    * e.g. if we ask for External Data, then switch to Workspace, do not render the results of the fetch for the External Data
* In addition, I audited our usage of Subject/. Subscription in many places and tried to ensure that there is an unsubscribe everywhere that it is needed
    * NOTE: we **only** need to call `unsubscribe()` for long-running data streams (e.g. subscriptions to syncService), but unsubscribe is not needed for one-off HTTP calls. This makes it very confusing to audit thoroughly and very easy to overlook, as it is quite situational.

## How to Test

#### Avalanching Requests
3. Navigate to Run > Files > External Data
4. Open the developer console
5. Click to Metadata, clear the Console, and click back to Files
    * Note the number of requests being sent here
6. Repeat step 5 and verify that the request count is the same, and has not increased

#### #261 File Manager sometimes shows wrong content
3. Navigate to Run > Files > External Data
4. Try really hard to break things - click around navs quickly, into subfolders, etc
    * Please make note of any odd behavior, and attempt as much as possible to reproduce it

### DIFF
--- a/src/app/+run-tale/run-tale/tale-files/tale-files.component.ts
+++ b/src/app/+run-tale/run-tale/tale-files/tale-files.component.ts
@@ -342,18 +342,22 @@ export class TaleFilesComponent implements OnInit, OnChanges {
         this.runService.runListRuns({ taleId: this.taleId })
                        .pipe(enterZone(this.zone))
                        .subscribe((r: Array<Run>) => {
-                         this.folders.next(r.sort(sortByUpdated));
-                         this.files.next([]);
-                         this.ref.detectChanges();
+                         if (this.currentNav === 'recorded_runs') {
+                           this.folders.next(r.sort(sortByUpdated));
+                           this.files.next([]);
+                           this.ref.detectChanges();
+                         }
                        });
         break;
       case 'tale_versions':
         this.versionService.versionListVersions({ taleId: this.taleId })
                            .pipe(enterZone(this.zone))
                            .subscribe((v: Array<Version>) => {
-                             this.folders.next(v.sort(sortByUpdated));
-                             this.files.next([]);
-                             this.ref.detectChanges();
+                             if (this.currentNav === 'tale_versions') {
+                               this.folders.next(v.sort(sortByUpdated));
+                               this.files.next([]);
+                               this.ref.detectChanges();
+                             }
                            });
         break;
       case 'home':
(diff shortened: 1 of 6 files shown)


src/app/+run-tale/run-tale/tale-files/tale-files.component.ts
```diff
@@ -163,6 +162,13 @@ export class TaleFilesComponent implements OnInit, OnChanges {
     this.load();
   }
 
+  ngOnDestroy(): void {
+    this.updateSubscription?.unsubscribe();
+
+    this.folders.unsubscribe();
+    this.files.unsubscribe();
```
bodom0015, line 169: Noting that I believe this is what actually fixed the duplicate requests shown by @Xarthisius 

Previously the `async` pipe used to automatically unsubscribe from observables, but this doesn't appear to be the case here anymore

src/app/+run-tale/run-tale/tale-metadata/tale-metadata.component.ts
```diff
@@ -127,6 +129,10 @@ export class TaleMetadataComponent implements OnInit {
     });
   }
 
+  ngOnDestroy(): void {
+    this.updateSubscription.unsubscribe();
+  }
```
bodom0015, line 134: Tale Metadata also had a rogue subscription

src/app/+run-tale/run-tale/tale-versions-panel/tale-versions-panel.component.ts
```diff
@@ -78,12 +78,6 @@ export class TaleVersionsPanelComponent implements OnInit, OnChanges, OnDestroy
         this.refresh();
       }
     });
-
-    this.versionsSubscription = this.syncService.taleUpdatedSubject.subscribe((taleId) => {
-      if (taleId === this.tale._id) {
-        this.refresh();
-      }
-    });
```
bodom0015, line 86: This is duplicated by the 6 lines directly preceeding it, removed the duplicate to prevent un-needed subscriptions from firing

Xarthisius:
I no longer see avalanching requests, but while switching back and forth between Files > Workspace (with 3 files and a folder) and Metadata tabs I get:

```
main.js:1 ERROR e {message: 'object unsubscribed', name: 'ObjectUnsubscribedError'}
```
in the console.

bodom0015:
This looks to be an innocuous error, but I am unclear what's going on here. I've confirmed that this problem only shows up in prod (when building with `--configuration production`), and can confirm that commenting out those two unsubscribe statements in `tale-files.component.ts` ngOnDestroy handler is the key.

I'm not sure what the error is here, since this appears to actually correct the bad behavior. I'll keep digging and see if I can figure out what's going on there :+1:

bodom0015:
This was a previous issue, so that warning is ok to ignore... what's happening is that there is a race condition being detected there. Noting that this behavior can happen with both the "Home" and "External Data", where we must do an external lookup in Girder before being able to lookup the folders/files present:
* Home does a lookup on the user to find their Home folder
* External Data does a lookup to find the Data Catalog

I can comment out those warnings, as they are less helpful now (but they were very helpful in detecting and debugging the race condition)

craig-willis:
The warning was apparently not the issue. This happens to me every time how:

* Create new tale
* External Data is empty
* Upload file to workspace
* Switch to External Data, workspace file shows.

https://user-images.githubusercontent.com/4334350/162969276-f822ff37-5914-4f15-8a9e-e9d21089bea2.mov

bodom0015:
Good catch! The following bit of logic was missing, which should clear out the External Data list when the user switches to `external_data` and the `tale.dataSet` is empty:
```typescript
        if (this.tale.dataSet.length === 0) {
          this.folders.next([]);
          this.files.next([]);
          this.loading = false;
          this.ref.detectChanges();

          return;
        }
```

Xarthisius (approved):

craig-willis (approved):
@Xarthisius identified my problem. I haven't seen any issues since registering.

PR #268, opened 2022-04-01
Allow user to upload entire folder structures

### LINKED ISSUE #233: Add support for uploading folders
This is something we've heard before and reinforced by one of our verification evaluation users.

**Problem**
The dashboard currently only allows users to upload files. If users have complex folder structures, they must manually create the structures and upload files per folder (or alternatively upload a single zipfile and unzip it using one of the IDEs).

**Proposed solution**
Enhance dashboard to allow users to upload a folder and all contents with a single selection.

bodom0015:
## Problem
It would be nice to be able to upload an entire folder at once. This is currently possible by zipping up the folder, uploading the zip, starting up a Tale, and using the Tale's console/terminal to unzip the folder via CLI. This is obviously a bit cumbersome for standard users, so we'd like to add a convenience method to easily enable people to upload an entire folder while maintaining its underlying structure

Fixes #233 

## Approach
Uses https://github.com/whole-tale/virtual_resources/pull/20

* Added new "folder upload" events to `file-browser` / `tale-files`
* Added a new file input with `directory` attributes: `<input type="file" directory webkit directory mozdirectory ... />`
* Added new `uploadFolder` method to `tale-files` - this ties in with our existing `uploadFiles`

## How to Test
Prerequisites: at least one Tale created, run with https://github.com/whole-tale/virtual_resources/pull/20

4. Navigate to Run > Files > Home for a Tale that you own
5. At the top-right, click on the plus (+) button
    * You should see a dropdown appear
    * You should see a new dropdown option: "Upload Folder..."
6. Click the "Upload Folder..." option
    * You should see a folder picker dialog pop up
    * You should notice that this folder picker dialog only allows you to select a folder
7. Choose a folder and click "Upload" at the bottom-right of the dialog
    * You should see the dialog close, and your uploads should begin
    * You should see the folder you selected now appears in the Files menu
8. Click down to navigate into the folder you uploaded
    * You should see all of the files from the folder that you chose are uploaded/uploading here

* Repeat steps for Run > Files > Tale Workspace

## TODOs
* ~~annoying scoping bug within `uploadFolder(filesToUpload)` where `filesToUpload` is appearing empty in rxjs callbacks~~
* ~~EXACERBATED ISSUE: multiple uploads bog down the backend, so we may need to throttle the number of active uploads in the queue~~ uploads now happen one at a time
* ~~KNOWN ISSUE: in-progress uploads do not properly render a progress bar or updates after folder navigation (move upload queue to notifications? pop up file queue?)~~ in-progress uploads should now show progress, even after navigating to subfolders
* KNOWN ISSUE: a hard refresh of the page will disrupt any in-progress downloads (not sure that we can get around this)
* ~~drag+drop functionality is still missing, if that is still desirable~~


bodom0015:
I've adjusted things so that it will only upload one file at a time, so that should prevent the frontend from sending too many simultaneous uploads :+1:

One open question is should we init all of the uploads at the start beforehand? Or is it ok to just init the uploads one at a time right before we send the bytes? Currently, the behavior is to init the upload right before sending the bytes, but I could be convinced to change this if there is good reason to pre-init them.

My thinking is that pre-init leaves behind blank `0B` file entries if any of the uploads fail or are aborted (e.g. user refreshes), but in-time creation might confuse the user if they don't immediately see the new uploads reflected in the file list

craig-willis:
Observed two problems:

1) Zero byte files are not uploaded and remain in pending state

2) When uploading multiple folders in a row, previous folders are re-uploaded. See video.
* Create three folders folder1, folder2, folder3
* Upload folder1, confirm uploaded
* Remove folder1, confirm removed
* Upload folder2. Both folder1 and folder2 are uploaded
* Remove folder1 and folder2, confirm removed.
* Upload folder3. folder1, folder2, folder3 are uploaded...

https://user-images.githubusercontent.com/4334350/165347130-64215e4d-441e-4ecb-87e7-ec7f50b71c34.mov

bodom0015:
Updated this to properly indicate progress updates for in-progress uploads into subdirectories via this new feature. The key here was to use the `parentId` and `name` of the Upload (returned from POST /api/v1/file[/chunk]) and match these values with the `folderId` and `name` of the Item that we list in the file browser :+1: 

I believe this also addresses the case of zero-byte files, but despite initially having some ideas of why it might be occurring I have been unable to reproduce the multi-folder upload issue. This may have somehow been fixed alongside the merge conflicts or the progress indicator issues here.

While I didn't confirm by mounting into a running Tale, here is what I experienced when running a similar test case: http://recordit.co/dvn7Gvwu0p

Could there be some special bit of your original test case that I might be missing that could lead to us seeing different behavior? If you're still able to reproduce the same issue with the updated branch, I can take another look to see what might be going on there

Xarthisius:
I still have some problem that I thought were related to 0-bytes file, but now I'm not sure. I'm attaching the test case this time:
[6227a1e9640c8da918f96914.zip](https://github.com/whole-tale/ngx-dashboard/files/8613079/6227a1e9640c8da918f96914.zip)

Unzip ^^ and try uploading folder `6227a1e9640c8da918f96914`. For me it stops on `fetch.txt` (which is 0-bytes) and in the console I see:

```
main.js:1 ERROR Error: Uncaught (in promise): Error: Syntax error, unrecognized expression: #upload-wtlocal:L3RtcC9kYXRhL3dvcmtzcGFjZXMvNi82MjcxNDZhMWIyMWU3NGRkZDhiZGVjNWYvNjIyN2ExZTk2NDBjOGRhOTE4Zjk2OTE0L2ZldGNoLnR4dHw2MjcxNDZhMWIyMWU3NGRkZDhiZGVjNjI=
Error: Syntax error, unrecognized expression: #upload-wtlocal:L3RtcC9kYXRhL3dvcmtzcGFjZXMvNi82MjcxNDZhMWIyMWU3NGRkZDhiZGVjNWYvNjIyN2ExZTk2NDBjOGRhOTE4Zjk2OTE0L2ZldGNoLnR4dHw2MjcxNDZhMWIyMWU3NGRkZDhiZGVjNjI=
    at Function.se.error (scripts.js:1:13478)
    at se.tokenize (scripts.js:1:21418)
    at se.select (scripts.js:1:22242)
    at Function.se [as find] (scripts.js:1:7046)
    at w.fn.init.find (scripts.js:1:24801)
    at new w.fn.init (scripts.js:1:25290)
    at w (scripts.js:1:1024)
    at e.<anonymous> (550.js:1:112490)
    at main.js:1:1845576
    at Object.next (main.js:1:1845681)
    at O (polyfills.js:1:17276)
    at polyfills.js:1:16380
    at a (main.js:1:1844441)
    at e.value (polyfills.js:1:7406)
    at Object.onInvoke (main.js:1:333578)
    at e.value (polyfills.js:1:7346)
    at t.value (polyfills.js:1:1522)
    at polyfills.js:1:18061
    at e.value (polyfills.js:1:8096)
    at Object.onInvokeTask (main.js:1:333388)
```

bodom0015:
Since there is now only one upload in progress, we don't need the progress bar's ID to be unique per-upload anymore. So I went ahead and changed the progressbar ID to be much more simple (no longer contains girder or virtual IDs to enforce uniqueness).

This way, it is much less error prone when updating the progress, but this may need to be revisited if we change our upload behavior in the future :+1:

Please let me know if you notice any other issues that might need to be addressed before this can be merged

Xarthisius (approved):

craig-willis (approved):
Sorry for the delay approving this.  Looks good. I've confirmed basic operation in Firefox and Safari as well.

PR #277, opened 2022-06-06
Public WholeTale Dashboard access as anonymous user

bodom0015:
## Problem
It would be nice to offer a read-only version of the Tale Catalog / Run Tale views. This would allow new users to preview the system before trying it out.

## Approach
* Adjust AuthGuard to allow unauthed users in Catalog/Run views
* Add a new button to the Login view to **Preview WholeTale** (note: button location/styling is subject to change)
* Disable some of the features that unauthed users can't use
    * Hide Shared/My Tales tabs on Catalog
    * Hide "Duplicate Tale" dropdown item
    * Hide Notification Stream
    * Hide link to Settings page
    * Hide Home Folder
    * Disable Run Tale button (w/ tooltip)

Screen Shot 2022-06-24 at 5 21 39 PM

Screen Shot 2022-06-06 at 2 35 26 PM

Screen Shot 2022-06-06 at 2 36 29 PM

## How to Test
Prerequisites: at least one Tale created and marked as "Public" 

2. Close all incognito windows, then open a new incognito window
3. Navigate to the WholeTale Dashboard
    * You should see a Read-Only version of the Public Tales Catalog, even without needing to log in
    * You should see that the "Run Tale button has been hidden on each Tale
    * You should see a "Sign In with Globus" button at the top-right
4. Click the title of the Tale or hover over the image and click the View button
    * You should be brought to a Read-Only version of the Run Tale view
    * You should see that the Run Tale button has been disabled (but offers a tooltip informing the user that they need to log in)
    * You should see that the "Interact" and "Share" tabs are not shown
6. Click the Files tab
    * You should see a Read-Only version of the file-explorer
    * You should see that the "Home" folder nav is not shown
7. Expand the dropdown menu at the top-right in the tale banner
    * You should see only "Read the docs" listed here

## TODOs
* Consistency between "Run Tale" button on Catalog view and on Run view
* ~~GET version and recorded runs are both returning a 401 when not logged in.. does this make sense?~~


bodom0015:
It looks like Tale Versions and Recorded Runs endpoints will throw a 401 if we’re not logged in? This probably makes sense for Recorded Runs, but are shouldn't the versions be public if the tale is marked as public?

Or perhaps are both of these resources always private and I should hide those navs as well?

craig-willis:
After rebuilding and clearing cache in an incognito session I'm getting the following error when simply loading the landing page:

```
ERROR Error: Uncaught (in promise): Error: Invalid CanActivate guard
Error: Invalid CanActivate guard
    at main.js:1:738207
    at e._subscribe (main.js:1:866534)
    at e.value (main.js:1:851005)
    at e.value (main.js:1:850789)
    at e.value (main.js:1:888198)
    at e.value (main.js:1:850667)
    at h (main.js:1:862753)
    at n.value (main.js:1:881429)
    at n.value (main.js:1:881316)
    at n.value (main.js:1:881137)
    at O (polyfills.js:1:17276)
    at O (polyfills.js:1:16811)
    at polyfills.js:1:18098
    at e.value (polyfills.js:1:8096)
    at Object.onInvokeTask (main.js:1:333414)
    at e.value (polyfills.js:1:8017)
    at t.value (polyfills.js:1:2310)
    at _ (polyfills.js:1:10387)
```

craig-willis:
Aside from the error, this is looking really good. 

Additional thoughts after preliminary review:
* The 401 on versions and runs doesn't make sense to me. If I login as a regular user, I can see versions and runs for read-only tales.
* With anonymous user support, I don't think we need the landing page. Perhaps there could be a "Sign In" link in place of the user menu?  If we have both Globus and Github, we'll still need some sort of sign-in page -- perhaps the TOS could go there?  We can move the acknowledgements elsewhere.
* Possibly indicate that the user needs to sign-in to do other things. Could be a message at the top of the content area or at least a tooltip over the "Run Tale" button?

When logged in as myself, I'm also seeing some unexpected behavior:
* After login, I only see the "Public Tales" tab (similar to anonymous user). If I select the "Tale Dashboard" link I see all three tabs (My/Shared/Public)
Screen Shot 2022-06-08 at 5 30 58 AM
* "Interact" tab,  "Run Tale" button and "Duplicate Tale" menu item are disabled. These are enabled on prod for me as a read-only user.

bodom0015:
I've updated this PR to address the sync issues and the CanActivate guard.

With the AuthGuard.CanActivate issue, we previously were guarding an entire route using `canActivate`. Because we are now guarding only particular child routes, we need to user `canActivateChild` in those cases instead. This requires a separate but similar interface to be implemented called `canActivateChild`, which we need to specify on the child routes instead.

The sync issue was something much more subtle and annoying with Angular module imports. In Angular 1.x, services were pure singletons - so this pattern was used to create a `ApiModule.TokenService` to hold the currently logged-in user/token and to share it between multiple components. Later versions of Angular provide much more control over how these services behave (which I clearly don't fully understand). Due to the way I was importing/nesting the modules here, it appeared that we had 2 separate ApiModules imported - e.g. setting a value in the TokenService for one instance wouldn't necessarily populate the same value in the other instance.

Previously:
```
AppModule => TaleCatalogModule => SharedModule => ApiModule => TokenService
          => RunTaleModule     => SharedModule => ApiModule => TokenService
```

New pattern:
```
AppModule => TaleCatalogModule => SharedModule
          => RunTaleModule     => SharedModule
          => ApiModule => TokenService
```

Importing ApiModule from AppModule (aka at the root level) and making sure we don't also import it within the child modules allows us to ensure that ApiModule and TokenService are only loaded once and that this singleton instance will be reused throughout.

Since these modules moved around, some dependencies may accidentally be broken. I played around a bit to make sure that common functionality for dependencies is still there (e.g. forms, modals, other stuff from material ui or semantic ui, etc), but let me know if you notice any other odd behavior there

craig-willis:
Looking good.

I haven't tested exhaustively, but noticed two things:
* AinWT isn't working for either auth'd or unauth'd users. For example, the following redirects to /public with no indication of unauthorized access for either user.
  * https://dashboard.local.wholetale.org/mine?uri=https%3A%2F%2Fsandbox.zenodo.org%2Frecord%2F105944
* Redirect to login isn't working when unauth'd and trying to access a restricted path:
  * Login as a valid user, create a tale, copy a URL
  * Logout and try to access the tale URL
  * Expected?: Redirect to login and I can access the tale.  This should probably be the case for any protected path -- or at least a subset?

craig-willis:
Testing AinWT specifically, I noticed the following:
* Route to `mine?url=` is preserved for a non-logged in user but the create tale modal does not display for me on AinWT for a logged in user.
* Route to `public?url=` works for both logged in and non-logged in users.

I think we should settle on a single route for AinWT. We selected `mine?url=` I expect because it worked. Would it make sense to have a specific route for AinWT -- e.g., `create?url=`?  With the current implementation, I guess I'd go with `public?url=` because it's working for both cases.

On an unrelated note, I noticed something odd with the Z-index for the notification panel:
Screen Shot 2022-06-30 at 6 49 59 PM

craig-willis (commented):
After additional testing, a could of issues:
* I've tried to address the versions/runs 401 with https://github.com/whole-tale/wt_versioning/pull/45. Even with this PR, the versions/runs don't display in the timeline (for a non-logged-in user) because of https://github.com/whole-tale/ngx-dashboard/blob/5110cf959501110dd85952a0229a760cd1482156/src/app/%2Brun-tale/run-tale/tale-versions-panel/tale-versions-panel.component.ts#L94.  Removing this line, the Export and Restore buttons appear in the version menu.
* For AinWT (assuming `public?uri=`), `asTale=false` isn't respected.    
( e.g., https://dashboard.local.wholetale.org/public?name=MyTale&uri=doi%3A10.7910%2FDVN%2FTJCLKP&asTale=false checks the READ/WRITE radio)

Xarthisius:
I run into an issue with the notification stream. If you go to https://dashboard.local.wholetale.org/ as an unauthenticated user it initially does three calls: `GET /tale`, `GET /user/me` and `GET /notification/stream`. The latter fails with 401:

bug

which is fine. However, if you login (and I think the condition is your browser is already authn with Globus, so there's no bouncing around OAuth pages) that connection is never re-established. As a result there are no visible notifications when you i.e. start a Tale. Also things like state change depending on whether instance is running or not are not happening cause there are no wt_notifications.

NOTE: There's something fishy going on with Chrome when you try to reproduce this bug. It worked first 3 times, but then I logged in and reloaded the page to establish proper /stream connection and for some reason it was no longer reproducible. Even if I closed Chrome, or used incognito session. That forced me to use Firefox.

bodom0015:
I've reworked a lot of the auth-guard and redirect logic here to (hopefully) be more flexible. 
It may be helpful to enumerate the ways that I'm testing this, in case I'm missing some of the various edge cases here?

The test URL I've been using to test the AiWT edge cases is here:
http://localhost:4200/mine?name=MyTale&uri=doi:10.7910%2FDVN%2FTJCLKP&asTale=false

I've tested navigating to this URL from the following states, and everything appears to work well enough:
* Logged out user -> Login -> AiWT
* Logged in user -> AiWT

Oddly, the UI navigates me to `/public` after login, but I still get the AiWT modal popup. This may be as close as we can get to properly handling things here for now.

In addition, fixed the following:
* Merged PR from @craig-willis into this branch to get the fixes for versions and runs and add the tooltip
* Fixed the z-index issue with notification stream (reported by @craig-willis) 
* Notification stream should attempt to reconnect when token changes (reported by @Xarthisius) 
* Fixed `asTale` querystring param handling (reported by @craig-willis) - likely unrelated to this PR, but this parameter was not being parsed properly as a string, causing that `asTale=false` to be evaluate to true (because it's a non-empty string)

craig-willis:
Noticing a few issues.  
* `/run` route broken: I can no longer access `/run` routes as either an anonymous or logged in user. I think this is because of https://github.com/whole-tale/ngx-dashboard/pull/277/files#diff-afdb4aa655be3c7396de2ac85e49ad6af90620a595a7898cef9a769b261c91cdR45-R49. If I change to `path: '',` it seems to work as expected (anon user can access public tale, logged in user can access private tale, `/` redirects to `/public` otherwise).
* Notifications: If I login and immediately try to run a tale, notifications do not appear without refresh.
* Edge case 1: with the above change, anon user accessing a private tale via direct URL gets a 401 + stack trace in the console. This could either redirect to sign-in or public?
* Edge case 2: Logging out while viewing a private tale leaves me on the tale page. Perhaps should redirect to `/public`?

Perhaps we could fix just the first item on this list, merge this and treat anything else as separate issues?

Xarthisius:
I run with a patch that Craig suggested:

```
diff --git a/src/app/+tale-catalog/tale-catalog.routes.ts b/src/app/+tale-catalog/tale-catalog.routes.ts
index 71d4662..7d1d646 100644
--- a/src/app/+tale-catalog/tale-catalog.routes.ts
+++ b/src/app/+tale-catalog/tale-catalog.routes.ts
@@ -43,7 +43,7 @@ export const routes = [
         },
       },
       {
-        path: '**',
+        path: '',
         pathMatch: 'full',
         redirectTo: 'public'
       },
```

and only hit the notification issue again. Otherwise it's looks good to me!

craig-willis:
Another edge case and possible side-effect of that patch:
* As anonymous user, go to https://dashboard.local.wholetale.org/
* Sign-in 
* Note that signed-in user is redirected to `/public`. Should be `/mine`?

Xarthisius (approved):

craig-willis (approved):

bodom0015:
Included the fix for the `/run` route precedence/ordering.. Good catch there :+1:

I also looked into the notification issue, but was unable to figure out why that connection is not being established. I can think of a way to fix it that would involve reworking things with [NgRx](https://ngrx.io/guide/store) (the Angular equivalent of React's [Redux](https://react-redux.js.org/introduction/getting-started)) now that I better understand how it works. That may be overkill here and would eventually involve sunsetting quite a few bits of related code (like SyncService). I'll think about this and see if there might be an easier way to get what we need for now, or if I should just bite the bullet and start moving things into a real state management system

In the meantime I will merge this and file these as new issues, with the notification issue taking higher priority

PR #278, opened 2022-06-16
Make "search" case insensitive and include category field

craig-willis:
## Problem
Searching for tales is currently case sensitive and doesn't include the category field. This fixes that.

## Approach
Update the `SearchTalesPipe` to match existing fields irrespective of case and add category as a searched field.

## How to Test
2. Create two tales: 1) Title: ABC 123 Category: XYZ and 2) Title: abc 123 Category: efg
3. Search for "abc" both tales should result; search for "xyz" and the first tale should appear

### Note:
I was surprised to learn that we implemented a Javascript-based search instead of using the  GET `tales` endpoint's text search feature. But this should work for now.

### DIFF
--- a/src/app/+tale-catalog/tale-catalog/pipes/search-tales.pipe.ts
+++ b/src/app/+tale-catalog/tale-catalog/pipes/search-tales.pipe.ts
@@ -14,17 +14,22 @@ export class SearchTalesPipe implements PipeTransform {
       return value;
     }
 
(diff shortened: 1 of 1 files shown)


bodom0015 (approved):
LGTM :+1: works as advertised

Would it be worth me spending some additional time to adjust this search to use the `text` field of `GET /tales? This might allow for easier paging and other user experience / performance features

PR #279, opened 2022-06-24
Handle error responses from registration/ainwt

Xarthisius:
## Problem

Errors raised during a) data registration and b) Tale creation are silently ignored.

## Approach

Create an error modal if something goes wrong

## How to Test

1. Follow https://github.com/whole-tale/girder_wholetale/pull/543 but simply do not provide ICPSR credentials
2. AinWT should fail and show modal
3. Create a Tale
4. Register doi:10.3886/E132081V1. Should fail and show modal.

### DIFF
--- a/src/app/+run-tale/run-tale/tale-files/tale-files.component.ts
+++ b/src/app/+run-tale/run-tale/tale-files/tale-files.component.ts
@@ -1,7 +1,7 @@
 import { ChangeDetectorRef, Component, EventEmitter, Input, NgZone, OnChanges, OnDestroy, OnInit, Output } from '@angular/core';
 import { MatDialog, MatDialogConfig } from '@angular/material/dialog';
 import { ActivatedRoute, Router } from '@angular/router';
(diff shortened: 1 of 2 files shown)


craig-willis (approved):
This works when I don't provide credentials, but if I provide an invalid password registration fails with. Not sure if we want to handle this.

```
AttributeError: AttributeError("'NoneType' object has no attribute 'split'")
[<FrameSummary file /girder/plugins/wholetale/server/tasks/import_binder.py, line 133 in run>, <FrameSummary file /girder/plugins/wholetale/server/lib/__init__.py, line 116 in register_dataMap>, <FrameSummary file /girder/plugins/wholetale/server/lib/import_providers.py, line 86 in register>, <FrameSummary file /girder/plugins/wholetale/server/lib/openicpsr/provider.py, line 246 in _listRecursive>, <FrameSummary file /girder/plugins/wholetale/server/lib/openicpsr/provider.py, line 226 in _get_payload>]
```

PR #280, opened 2022-06-27
feat: copyTale should copy specific Tale version

### LINKED ISSUE #267: Add ability to create a new tale based on a version
**Feature request:**
Add the ability to open a version of a tale as a new tale. 

**Source**:

From https://wholetale.slack.com/archives/C02FQLTR67P/p1648577005526759:
> Out of curiosity: Saved versions of the Tale can be renamed/deleted/downloaded. Why is "open in a new tale" not an option? It would seem a natural step (emulating the export-to-zenodo-then-re-import-to-WT functionality).
> The functionality is there, just nobody asked for it and wanted a button
> Can I ask for it? That makes n=1...

bodom0015:
## Problem
~~`copyTale` should copy the specific version that the user is viewing~~
The UI should offer a way to clone a specific Tale version as a new Tale, without including all other/previous version info.

Fixes #267

## Approach
* ~~Send `versionId` along with the API request to `/tale/:id/copy`~~
* Add a new dropdown item for Tale Versions: "As New Tale"

## How to Test
Prerequisites: at least one Tale created, at least two versions created

2. Log into the WholeTale Dashboard
3. Create a Tale (if you haven't already)
4. On the Tale, create two distinct versions (if you haven't already)
5. Navigate to Run > Files > Tale Versions for the Tale
    * You should see your 2 versions listed here
6. Expand the dropdown beside one of the Versions and choose "As New Tale"
    * You should be redirected to the Run > Files view for your newly-created Tale copy
    * You should no longer see the other versions listed here (beside the one that was selected for "As New Tale" originally)

### DIFF
--- a/src/app/+run-tale/run-tale/run-tale.component.ts
+++ b/src/app/+run-tale/run-tale/run-tale.component.ts
@@ -379,11 +379,10 @@ export class RunTaleComponent extends BaseComponent implements OnInit, OnChanges
     }
 
     copyTale(): void {
-      this.taleService.taleCopyTale(this.tale._id, this.tale.restoredFrom).subscribe(res => {
-        this.logger.debug("Tale copying:", res);
+      this.taleService.taleCopyTale(this.tale._id).subscribe(res => {
+        this.logger.debug("Cloning entire Tale (including versions):", res);
       });
     }
-
 
     // Expected parameter format:
     //    dataMap: [{"name":"Elevation per SASAP region and Hydrolic Unit (HUC8) boundary for Alaskan watersheds","dataId":"resource_map_doi:10.5063/F1Z60M87","repository":"DataONE","doi":"10.5063/F1Z60M87","size":10293583}]
--- a/src/app/+run-tale/run-tale/run-tale.component.ts
+++ b/src/app/+run-tale/run-tale/run-tale.component.ts
@@ -337,8 +337,9 @@ export class RunTaleComponent extends BaseComponent implements OnInit, OnChanges
 
     saveTaleVersion(): void {
       this.logger.debug('Saving Tale version');
-      this.versionService.versionCreateVersion({ taleId: this.taleId, force: true }).subscribe(version => {
+      this.versionService.versionCreateVersion({ taleId: this.taleId, force: true }).subscribe((version: Version) => {
         this.logger.debug("Version saved successfully:", version);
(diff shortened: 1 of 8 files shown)


craig-willis:
As discussed on the 8/1/2022 dev call, my vague issue caused some confusion on this one. What the user was proposing is a way to create a new tale from a previous version.

We discussed the following regarding this PR:
* The top-level "Duplicate Tale" should behave as it does today, duplicating the entire tale including all versions and runs.
* We will add a new menu item on the File Manager > Saved Versions context menu that allows the user to create a new tale from the selected version. The proposed menu item is "As Tale". This would call the copy endpoint with the version ID, resulting in a new tale based on the selected version.

From my understanding, the use case for this is to be able interactively run and compare two different versions of a tale at the same time.

bodom0015:
Updated this PR per our discussions yesterday :+1: on RunTale dropdown, `copyTale` once again copies the entire Tale (including all existing version info)

I've added a new dropdown item to Run > File > Versions to clone "As New Tale", and updated FileService to include the new `shallow` parameter. I am sending that `shallow` parameter (e.g. POST to `https://girder.local.wholetale.org/api/v1/tale/62e97ed6bec07a0cf441373d/copy?versionId=62e97ed6bec07a0cf4413745&shallow=true`). This seems to copy only the desired version :+1:

bodom0015:
I am seeing some sync issues with this:

1. Creating a new Tale Version does not automatically refresh the Files view (I think this is a known issue), so creating new versions on the right can lead to confusion and desync on the left
2. Simply navigating to the new Tale does not cause the Tale Versions Panel to refresh.. so the user can still see the stale version timeline listed here

Xarthisius:
LGTM! Can we add 'As New Tale' version specific dropdown in "Version History" panel? Unless we discussed we why shouldn't and I just don't remember :)

Xarthisius (approved):

bodom0015:
I think we discussed consolidating stuff between the Tale Versions panel and the File Browser, but I'm not sure where we landed on that one. I can add that to the Versions Panel for now as well, until we can figure out where we want it to live in the long term

I also was unable to reproduce #287 (new closed), and pushed a small fix for #288 :+1:

craig-willis (approved):
Works for me. 

If I recall, we decided to move most of the operations from the Tale History panel to the File Browser instead of duplicating, since Tale History links to the file browser. I'll create a separate issue for discussion.

PR #286, opened 2022-08-02
fix: prevent user from clicking Add multiple times in Run > Share

### LINKED ISSUE #266: Duplicated entries in Tale's access created by an accidental double click
**Steps to Reproduce**

1. Create Tale
2. Navigate to Tale > Share. Select a random other user.
3. Double click on "Add"

**Expected Results**

1. Other user is added only once.

**Actual Results**

wt

bodom0015:
## Problem
On the Run > Share view, if the user clicks Add multiple times in succession then they can end up with duplicate entires in the access list.

Fixes #266 

## Approach
Disable the "Add" button until we are finished processing the current request.

## How to Test
Prerequisites: at least one Tale that you own

3. Navigate to Run > Share for a Tale that you own
4. Search for a user (e.g. `John Doe`)
5. Quickly attempt to press Add multiple times
    * You should be unable to click Add multiple times, as it should disable while processing the current request

### DIFF
--- a/src/app/+run-tale/run-tale/tale-sharing/tale-sharing.component.html
+++ b/src/app/+run-tale/run-tale/tale-sharing/tale-sharing.component.html
@@ -57,7 +57,7 @@
                 <div class="item" data-value="2">Owns</div>
               </div>
            </div>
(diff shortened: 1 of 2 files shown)


craig-willis (approved):
wfm

Xarthisius (approved):

PR #291, opened 2022-08-16
fix: notifications should appear without refresh, fix login/logout edge cases for private Tale

### LINKED ISSUE #283: Notifications do not appear without refresh
From #277:
> Notifications: If I login and immediately try to run a tale, notifications do not appear without refresh.

**Steps to Reproduce**
Prerequisites: at least one Public Tale, Tale is not running

1. Navigate to WholeTale UI as a logged-out user
2. Log in to the platform
3. Without refreshing or navigating, Launch a Tale

**Expected Results**
The notification stream should open, and let you know the status of the launching Tale

**Actual Results**
The notification stream does not display

**NOTES**
Possible fixes might include:
* force a refresh after successful login to ensure that the UI inits with the correct state
* actual state management with NgRX - note this would be a larger undertaking, as it may affect other unrelated code (for example, using NgRx would totally negate the need for SyncService)

### LINKED ISSUE #284: silent 401 in the console when accessing private Tale
From #277:
> Edge case 1: with the above change, anon user accessing a private tale via direct URL gets a 401 + stack trace in the console. This could either redirect to sign-in or public?

**Steps to Reproduce**
Prerequisites: at least one private Tale

1. Navigate to /run for a private Tale as a logged-out user

**Expected Results**
Redirected to sign-in

**Actual Results**
401 in the console, no error message displayed to the user, user is routed to `/public`

### LINKED ISSUE #285: Logout while on private Tale should redirect to public
From #277:
> Edge case 2: Logging out while viewing a private tale leaves me on the tale page. Perhaps should redirect to `/public`?

Similar to #284 (but in reverse)

**Steps to Reproduce**
Prerequisites: at least one private Tale

1. Navigate to `/run` for a private Tale as a logged-in user with sufficient access
2. Log out of the platform

**Expected Results**
User should be redirected to `/public`

**Actual Results**
User's location does not change and they can still see the private Tale to which they should no longer have access

**NOTES**
Possible fixes might include:
* force a refresh after successful login to ensure that the UI inits with the correct state
* actual state management with NgRX - note this will be a larger undertaking, as it may affect other unrelated code (for example, using NgRx would totally negate the need for SyncService)

bodom0015:
## Problem
Some remaining problems still linger after merging #277 and #282. Namely

Fixes #283 - `notification-stream` sometimes does not display when launching a Tale, events are not handled properly
Fixes #284 - silent 401 error in the console when attempting to access a private Tale before logging in
Fixes #285 - user can still view stale data after logging out when viewing a private Tale

## Approach

* #283 - Reconnect to `notification-stream` when user is logged in, disconnect on logout
* #284 - Re-route to login automatically if user attempts to access a private Tale
* #285 - Logout while on private Tale's Run view now redirects to `/public` after logout

## How to Test
Prerequisites: logged out, at least one Tale (private; not public, not shared)

2. Navigate directly to `/mine` (without logging in)
    * You should be redirected to the login flow (via Globus)
3. Follow the login flow until you're logged in
    * You should be brought to `/mine` on the WholeTale Dashboard
4. Without navigating or refreshing, Launch a Tale
    * #283 - You should see the `notification-stream` appears without needing to refresh the page
5. Wait for the Tale to launch
    * Once the Tale has started, you should see the Interact tab automatically re-renders without needing to refresh the page
6. Save the `/run/:id` URL from your address bar for later (see step 8)
7. At the top-right, expand the User dropdown and click Logout
    * #285 - You should be redirected to `/public`, since this Tale is private and you are no longer authorized to view it after logging out
8. Navigate back to the `/run/:id` view by directly pasting the URL you saved previously
    * #284 - You should be rerouted through the login flow and end up on the `/run/:id` view, now authorized to view it

### DIFF
--- a/angular.json
+++ b/angular.json
@@ -72,6 +72,10 @@
                 }
               ]
             },
+            "dev": {
+              "optimization": false,
+              "sourceMap": true
+            },
             "production": {
               "fileReplacements": [
                 {
@@ -100,6 +104,9 @@
             "hmr": {
               "hmr": true,
               "browserTarget": "universal:build:hmr"
+            },
+            "dev": {
+              "browserTarget": "universal:build:dev"
             },
             "production": {
               "browserTarget": "universal:build:production"
--- a/package.json
+++ b/package.json
@@ -14,7 +14,7 @@
   "scripts": {
     "ng": "ng",
     "clean": "rimraf .cache dist coverage test-report.xml",
-    "start": "ng serve",
+    "start": "ng serve --configuration=dev",
     "start:prod": "ng serve --configuration production",
     "build": "ng build",
     "build:prod": "ng build --configuration production",
--- a/src/app/+run-tale/run-tale/run-tale.component.ts
+++ b/src/app/+run-tale/run-tale/run-tale.component.ts
@@ -171,13 +171,6 @@ export class RunTaleComponent extends BaseComponent implements OnInit, OnChanges
       this.logger.debug(`Fetching tale with _id=${this.taleId}`);
       this.taleService.taleGetTale(this.taleId)
                       .subscribe((tale: Tale) => {
-        if (!tale) {
-          this.logger.error("Tale is null, something went horribly wrong");
-          this.router.navigate(['public']);
-
-          return;
-        }
-
         this.logger.info("Fetched tale:", tale);
         this.tale = tale;
 
--- a/src/app/+run-tale/run-tale/run-tale.component.ts
+++ b/src/app/+run-tale/run-tale/run-tale.component.ts
@@ -205,6 +205,12 @@ export class RunTaleComponent extends BaseComponent implements OnInit, OnChanges
       this.detectTaleId();
       this.detectCurrentTab();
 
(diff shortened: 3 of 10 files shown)


craig-willis (approved):
These are all working now for me, thanks.

One very minor tangential question/suggestion -- perhaps the sign-in link could redirect to `/mine` instead of `/public`?

bodom0015:
I can probably change this, but noting that Sign In simply keeps you on the same page. So if you're on `/public` when you click it, it will bring you to `/public` (same with `/run/:id`)

If desired I can try to override this behavior so that clicking Sign In on `/public` ends up dropping the user on `/mine` instead, but should note that this breaks consistency with the behavior we use everywhere else

craig-willis:
Thanks, @bodom0015. I didn't realize that was the behavior.  As a non-logged in user, I think I only have access to `/public` and `/run`, and certainly would want sign-in while accessing a tale to stay on the `/run` view.  I retract...

Xarthisius (approved):

PR #295, opened 2022-08-30
fix: shorten rd by omitting hostname

Xarthisius:
Companion to https://github.com/whole-tale/girder_wholetale/pull/552 . Shorten the redirect url by not hardcoding dashboard's hostname. 

## How to Test

1. Login to dashboard.
2. Create a Tale, make it public. Navigate to view / interact tab.
3. Copy the url.
4. Logout.
5. In an incognito session: open the url from 3.
6. You should end up being forced to login and afterwards on the proper interact tab.


craig-willis (approved):
WFM. An alternative test case is to just leave the tale private. Accessing any tab will route through login.

src/app/layout/header.component.ts
```diff
@@ -146,9 +146,18 @@ export class HeaderComponent extends BaseComponent implements OnInit, OnDestroy,
...
-          const route = rd.split(window.origin)[1];
-
-          return this.router.navigateByUrl(route);
+          if (rd.indexOf('?') > 0) {
+            const [path, query] = rd.split('?');
+            const searchParams = new URLSearchParams(`?${query}`);
+            const params = Object.fromEntries(searchParams.entries());
+
+            return this.router.navigate(
+              path.split('/').filter((n: any) => n),
```
bodom0015, line 155: Any reason for the call to `filter` here? This appears to be a no-op, but maybe I'm missing some subtle behavior here?

src/app/layout/header.component.ts
```diff
@@ -146,9 +146,18 @@ export class HeaderComponent extends BaseComponent implements OnInit, OnDestroy,
         // Redirect via querystring param, if provided
         const rd = this.route.snapshot.queryParams.rd;
         if (rd) {
-          const route = rd.split(window.origin)[1];
-
-          return this.router.navigateByUrl(route);
```
bodom0015, line 151: For the `rd` parameter, `route` was previously created by stripping off the origin/hostname from the front... isn't `pathname` + `search` the same thing as `route` used to be? Are there perhaps cases where `pathname + search` is not the same as `href - origin`?

Unless I'm confusing myself here (which is very possible), we should still be able to pass that into `navigateByUrl` without needing to parse it out manually

tsconfig.json
```diff
@@ -23,7 +23,7 @@
       "@layout/*": ["./src/app/layout/*"]
     },
     "typeRoots": ["node_modules/@types"],
-    "lib": ["es2017", "dom"]
+    "lib": ["es2019", "dom", "dom.iterable"]
```
bodom0015, line 26: FWIW I have no issue with including `dom.iterable` here, but I am also unclear why it was needed in this case.

Was there a build error/warning that you were hitting? Typically `DOM.Iterable` is needed if we want to loop over collections of UI elements from the DOM, which I don't see happening here.

For more info, see: https://errorsfixing.com/in-typescript-config-file-tsconfig-json-is-it-redundant-to-set-lib-compiler-option-to-both-dom-and-dom-iterable/

bodom0015 (approved):
LGTM! I did have a few minor questions, but overall looks good :D

src/app/layout/header.component.ts
Xarthisius:
```
"/foo/bar".split("/")
(3) ['', 'foo', 'bar']
"/foo/bar".split("/").filter(n => n)
(2) ['foo', 'bar']
```

tsconfig.json
Xarthisius: If you try to build without `dom.iterable` you'll get:

```
Error: src/app/layout/header.component.ts:152:60 - error TS2339: Property 'entries' does not exist on type 'URLSearchParams'.

152             const params = Object.fromEntries(searchParams.entries());
                                                               ~~~~~~~

```

src/app/layout/header.component.ts
Xarthisius: It wasn't working. We could see `https://dashboard.local.wholetale.org/` in `rd` during oauth flow. You can double check if that's the case on master.

src/app/layout/header.component.ts
bodom0015: Ahh.. so this is just to cover malformed URLs in the `rd`? We should never see a `//` in there, unless the user is manually messing with their `rd` parameter (which we don't support)

tsconfig.json
bodom0015: Ooooh, that's interesting.. I never viewed URLSearchParams as a DOM element, but if it fixed the error then it works for me :+1:

src/app/layout/header.component.ts
bodom0015: Previous behavior:
`rd=http:%2F%2Flocalhost:4200%2Fpublic%3Fasdf%3D1234`
Then `route` becomes `/public?asdf=1234` and we can navigate to that using `navigateByUrl`

Current behavior:
`rd=%2Fpublic%3Fasdf%3D1234`
Then `route` becomes `/public?asdf=1234` and we should be able to navigate to it directly?

I have no issue with removing the hostname/origin from the `rd` parameter, but the string manipulation does not seem necessary at all

The following works for me without any parsing or anything:
```typescript
        if (rd) {
          this.router.navigateByUrl(rd)
        }
```

Xarthisius: ha! I didn't know that's possible. I'll try it out.

bodom0015: :+1: if it does work, making this change would resolve my other comments as well

Xarthisius: Works like a charm! Thank you!!

PR #300, opened 2022-09-19
fix: create folder without refresh, fix dropdowns after rename folder

### LINKED ISSUE #297: Context menu doesn't display after rename
**Steps to Reproduce**
1. Create a new tale
2. Create folder "folder1"
3. Rename folder to "folder2"
4. Open context menu for folder

**Expected Results**

Context menu displays

**Actual Results**

Context menu doesn't display

**NOTES**
* Chrome Version 104.0.5112.101/MacOS 
* In the console I sometimes see an error "Transition: Element is no longer attached to DOM. Unable to animate"
* For me, sometimes step 2 folder doesn't appear without refresh
* This also happens for files

### LINKED ISSUE #299: Folder not displayed after create
**Steps to Reproduce**

* Sign in to https://dashboard.stage.wholetale.org
* Open dev console
* Create a new tale
* Go to files > home or workspace
* Create a folder
* Note no errors but new folder is not shown
* Refresh
* Note possibly-related console error  `TypeError: Cannot read properties of undefined (reading 'public')`
* Folder is displayed

**Expected Results**

Folder is displayed after successful creation

**Actual Results**

Folder is not displayed until refresh

**Notes**
* Maybe related to https://github.com/whole-tale/ngx-dashboard/issues/297
* Related/same problem with Move/Copy To Workspace (file is moved/copied but UI doesn't display)

bodom0015:
## Problem
Some issues with the file browser:
* #299: When user creates a folder it does not immediately appear in the file browser
* #297: When user renames a folder it breaks the dropdown menu of that folder

Fixes #299
Fixes #297

## Approach
* Call `this.load()` after folder creation, rather than manually adding the new folder to the list - this will force a full refresh of the contents of the file browser
* Call `this.initDropdowns()` in every possible callback of `load()` - we should rely on calling `this.load()` anytime the user makes a change to directory structure (e.g. copy, move, rename, upload, create, etc) and this should correctly re-init all of the dropdowns on the file browser if any DOM elements need to be re-rendered (which somehow breaks their dropdown menu)

## How to Test
Prerequisites: at least one Tale to which you have write access

3. Navigate to Run Tale > Files > Tale Workspace (or Home) for a Tale that you can modify
4. Create a folder
    * #299: You should see the folder appear almost immediately (without needing to manually refresh the page)
5. Expand the dropdown by clicking the [v] beside the folder you just created
    * You should see the dropdown appears as expected
7. Rename the folder you just created
    * You should see the name change almost immediately
8. Expand the dropdown once more by clicking the [v] 
    * #297: You should see the dropdown still appears as expected

### DIFF
--- a/src/app/+run-tale/run-tale/run-tale.component.ts
+++ b/src/app/+run-tale/run-tale/run-tale.component.ts
@@ -206,8 +206,8 @@ export class RunTaleComponent extends BaseComponent implements OnInit, OnChanges
       this.detectCurrentTab();
 
       this.tokenService.currentUser.subscribe((user: User) => {
-        if (!user && !this.tale.public) {
-          this.router.navigate(['public'])
+        if (!user && !this.tale?.public) {
+          this.router.navigate(['public']);
         }
       });
 
--- a/src/app/+run-tale/run-tale/tale-files/tale-files.component.ts
+++ b/src/app/+run-tale/run-tale/tale-files/tale-files.component.ts
@@ -383,6 +383,14 @@ export class TaleFilesComponent implements OnInit, OnChanges, OnDestroy {
     })();
   }
 
+  initDropdowns(): void {
+    setTimeout(() => {
+      $('.ui.file.dropdown').dropdown({ action: 'hide' });
+      this.load();
+    }, 500);
+
+  }
+
 
   load(): void {
     // Already loading, short-circuit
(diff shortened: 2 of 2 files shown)


Xarthisius:
Dropdown folder started to behave weird in points 5. and 7. It displays and immediately folds back before I even have a chance to move the mouse. It doesn't happen always, but more often than not.

bodom0015:
Good catch! Sorry, I had overlooked something stupid here where calling `this.load()` would call `this.initDropdowns()`, which would then call `this.load()` again, etc. So it was constantly fetching and re-rendering the file-browser in the background, and each time it would re-init the dropdowns and close them.

This should happen far less frequently now :+1:

craig-willis (approved):
wfm

Xarthisius (approved):

PR #301, opened 2022-09-22
fix: interact tab should route user through login if needed

### LINKED ISSUE #296: User not prompted to login when accessing interact tab for public tale.
**Steps to Reproduce**
Encountered when testing https://github.com/whole-tale/ngx-dashboard/pull/295
1. Login to dashboard.
1. Create a Tale, make it public. Navigate to view / interact tab.
1. Copy the url.
1. Logout.
1. In an incognito session: open the url from 3.

**Expected Results**
Forced to login and afterwards on the proper interact tab.

**Actual Results**
"You must Run this Tale before you can Interact" message appears.

**NOTES**
* I'm not 100% sure what the behavior should be here. Prompting to login makes sense, but so would a message that says "You must login to run this tale. As it is, we offer only a tooltip on the run button to tell the user they need to login.
* I noticed that `tab=sharing` briefly shows the Sharing tab content before redirecting to login.

bodom0015:
## Problem
An unauthenticated user can end up on the Interact tab, which does not work properly if the user is not logged in.

Fixes #296

## Approach
* Force users to login if they reach the Interact tab without doing so (note that Interact tab is hidden from UI if user is not logged in)
* Add a message to the Interact page that the user must Sign In to Interact - just in case the automatic redirect fails

## How to Test
Prerequisites: at least one public Tale

4. Mark the Tale as public
5. Navigate to the Interact tab and copy the URL
6. Sign out and open a new incognito window
7. Paste the Interact URL that you saved previously
    * You should be prompted to log in
8. Enter your credentials (2FA, etc) and sign in
    * You should end up on the Interact tab, as an authenticated user

### DIFF
--- a/src/app/+run-tale/run-tale/run-tale.component.ts
+++ b/src/app/+run-tale/run-tale/run-tale.component.ts
@@ -176,6 +176,12 @@ export class RunTaleComponent extends BaseComponent implements OnInit, OnChanges
 
         this.refreshCollaborators();
 
+        this.tokenService.currentUser.subscribe((user: User) => {
+          if (!user && !this.tale.public) {
+            this.router.navigate(['']);
+          }
+        });
+
         this.userService.userGetUser(this.tale.creatorId)
                       .subscribe(creator => {
           this.creator = creator;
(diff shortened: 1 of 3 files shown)


craig-willis (approved):
This works for me.  I'm prompted to login as expected and haven't found a way to see the message.

Xarthisius (approved):

PR #303, opened 2022-09-30
feat: move TaleVersionPanel dropdown options to TaleFilesTab

### LINKED ISSUE #294: Consolidate version operations
As discussed in https://github.com/whole-tale/ngx-dashboard/pull/280, there are inconsistencies between the File Browser and Tale History panel menus for versions. I believe we discussed consolidating all of these on the File Browser, since the Tale History links there for a version:

Today: 
* Tale History:
  * Version: View Info, Restore, Export; Link to File Browser for Version
  * Recorded Run: View Info; Link to File Browser for Version and Run
* File Browser: 
  *  Version: Rename, Remove, Download, As New Tale
  *  Recorded Run: Rename, Remove, Download

I propose that we move all operations from the Tale History menu to the File Browser, possibly even removing the menu from the Tale History panel altogether. Users have also reported confusion between "Download" and "Export" for versions. I suggest that we rename this to "Download Folder" for (maybe) some clarity. Also, maybe we can link from the Recorded Run to the Version or offer a menu to Export the associated version (e.g., "Export Version" for both runs and versions).

Proposed:
* Tale History: Remove context menu
* File Browser:
  * Version: View Info, Rename, Restore, Download Folder, Export Version, As New Tale, Remove 
  * Recorded Run: View Info, View Version (?), Rename, Download Folder, Remove

bodom0015:
## Problem
FileBrowser and TaleVersionsPanel have a mismatch between the dropdown option available

Fixes #294 

## Approach
* Move the dropdown functions that are currently available on the TaleVersionsPanel to the TaleFilesTab / FileBrowser
* NOTE: New dropdown options are currently only available if user is viewing the list of versions, but we can look into possibly offering them lower in the folder hierarchy if it might be more intuitive and/or user-friendly

Available dropdown options for each case are enumerated below.

Tale Versions Panel (no dropdown, styling is improved in #302):
Screen Shot 2022-09-30 at 11 52 54 AM

Run > Files > Tale Versions dropdown options (at the root):
Screen Shot 2022-09-30 at 11 54 23 AM

Run > Files > Tale Versions dropdown options (in a subfolder):
Screen Shot 2022-09-30 at 12 05 25 PM

Run > Files > Recorded Runs dropdown options (at the root):
Screen Shot 2022-09-30 at 12 04 13 PM

Run > Files > Recorded Runs dropdown options (in a subfolder):
Screen Shot 2022-09-30 at 12 05 01 PM

## How to Test

Prerequisites: 

3. ...


Xarthisius:
@bodom0015 could you please fix the conflict?

bodom0015:
Conflicts should be fixed.. lots of it was just an indentation change (wrapped in a new `div`) :+1:
not sure if you wanted to retest here, but will leave this just in case :pray:

Xarthisius:
LGTM! I run into small CSS issue where action dropdown is truncated (though you can scroll to get the rest of it), but I'm not sure if it's directly related to this PR:

dashboard

bodom0015:
I've pushed a small fix for this CSS issue :+1: thank you!

Screen Shot 2022-11-07 at 1 18 34 PM

craig-willis:
This looks good, but I don't think Copy should be available for either versions or runs (it results in 500 anyway)

I noticed a few things that are either non UI or can be addressed in the future, but want to put them here for discussion:
* "Download as Folder" and "Export Version" produce two different zips (regular zip of version folder v bag of tale version) but they both have the same name. @Xarthisius thoughts?
* I wonder if we should add "Export Version" to the run dropdown. We've had feedback in the past that it's not clear how to export/publish a run (you can't, you can only export/publish a version). But it seems possible to just export the version here.
* I thought we had a way to publish a version in the past, but don't see it (and it wasn't in the issue I created). Seems like an easy add to the dropdown? Can be a separate PR in the future.

craig-willis (approved):

PR #304, opened 2022-10-17
fix: allow to use AinWT with git repo

Xarthisius:
## Problem

Currently it's not possible to initialize AinWT in a "git variant" by crafting an url to pointing to dashboard.

## Approach

1. Modify `CreateTaleModalComponent` to utilize `uri` query parameter if `mode.Git` was selected.
2. In case of AinWT detect whether `git=true/false` parameter was used and choose mode accordingly

## How to Test

2. Follow https://dashboard.local.wholetale.org/mine?name=AEAREP-3198-Stata&environment=STATA%20%28Desktop%29&uri=https%3A%2F%2Fgithub.com%2FAEADataEditor%2Freplication-template%2F&asTale=true&git=true
3. Git variant of 'create Tale' modal should be opened with prefilled data. Click on "Create a new Tale".
4. A Tale with git repo cloned into the workspace should be created.
5. (BEWARE) While I tested my changes work as intended, I haven't verified that a classic AinWT still works, i.e. I haven't broken all the other things... You may want to check that.

### DIFF
--- a/src/app/+tale-catalog/tale-catalog/modals/create-tale-modal/create-tale-modal.component.ts
+++ b/src/app/+tale-catalog/tale-catalog/modals/create-tale-modal/create-tale-modal.component.ts
@@ -7,7 +7,7 @@ import { LogService } from '@shared/core';
 // import * as $ from 'jquery';
 declare var $: any;
 
-enum Mode {
+export enum Mode {
   Default = "default",
   Git = "git",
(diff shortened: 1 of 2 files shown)


craig-willis (approved):
This works for me including additional AinWT tests.

bodom0015 (approved):
LGTM! Works as advertised :+1:

PR #310, opened 2022-12-05
feat: notification stream - add a Cancel button beside job progress bar

### LINKED ISSUE #307: Add UI component to cancel image build and recorded run jobs.
The Girder API `PUT /job/{id}/cancel` has always been there, but our jobs have not respected it until https://github.com/whole-tale/gwvolman/pull/177/files.

To allow users to cancel these jobs, we need to add a UI component. From my understanding, cancel is currently limited to these two job types. This will likely need to go into the notification panel, but there may be a better place.

bodom0015:
## Problem
There is no way to cancel running jobs from the WholeTale Dashboard

Fixes #307 

## Approach
* NotificationStream - add a Cancel button beside the progress bar of running jobs
    * Show this button only if the `event.data.state === 'active'`
    * Button should only be clickable if `event.data.resource.jobs.length` is truthy

Screen Shot 2022-12-05 at 1 08 29 PM

## How to Test

3. Launch a Tale
    * You should be brought to the Run Tale page
    * You should see the NotificationStream appear at the top-left with a new event
    * You should see that the new event shows that your Tale image is now building
    * You should see a red X button beside this running job - this is the Cancel button
4. Click the X beside the running Job
    * NOTE: you may need to click multiple times, as multiple jobs are sometimes triggered after one another (e.g. launch instance = build image + create volume + start container)
    * You should see that eventually the job goes into the Error state and the Cancel button is no longer clickable
5. Expand the Tale dropdown at the top-right, choose "Recorded Run", and Submit the modal dialog that appears
    * You should see the NotificationStream appear at the top-left with a new event
    * You should see that the new event shows that your Recorded Run is now executing
    * You should see a red X button beside this running job - this is the Cancel button
6. Click the X beside the running Job
    * NOTE: you may need to click multiple times, as multiple jobs are sometimes triggered after one another
    * You should see that eventually the job goes into the Error state and the Cancel button is no longer clickable

### DIFF
--- a/src/app/layout/notification-stream/notification-stream.component.html
+++ b/src/app/layout/notification-stream/notification-stream.component.html
@@ -50,6 +50,7 @@
       <td width="5%">
         <button class="ui tiny negative basic right floated icon button"
                 [disabled]="!event?.data?.resource?.jobs?.length"
+                matTooltip="Cancel"
                 *ngIf="event.data.state === 'active'"
                 (click)="cancelJob(event)">
           <i class="times icon"></i>
--- a/src/app/layout/notification-stream/notification-stream.component.html
+++ b/src/app/layout/notification-stream/notification-stream.component.html
@@ -34,6 +34,10 @@
           <i class="ui fas fa-fw fa-pause yellow icon"></i>
           Queued<span *ngIf="event.data.message">: {{ event.data.message }}</span>
         </span>
+        <span  *ngSwitchCase="'canceled'">
+          <i class="ui fas fa-fw fa-stop brown icon"></i>
+          Canceled<span *ngIf="event.data.message">: {{ event.data.message }}</span>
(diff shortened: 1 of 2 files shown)


craig-willis:
Overall this looks good. A couple of minor points:
* The big red X is pretty clear, but perhaps we could add a "Cancel" tooltip?
* I was hoping that we could change the 'Error' message to 'Canceled', but the jobs plugin doesn't seem to let us set the right state https://github.com/whole-tale/gwvolman/pull/177#issuecomment-1351573338
* I'm not sure what we can do about chained jobs -- I don't think they are all cancelable. Maybe there's a way to only display the cancel button for jobs that we know are cancelable?

craig-willis (approved):

Xarthisius (approved):

PR #311, opened 2022-12-05
feat: view Tale Instance logs

### LINKED ISSUE #308: Add UI component to allow user to view instance logs
https://github.com/whole-tale/girder_wholetale/pull/557 (see also https://github.com/whole-tale/deploy-dev/pull/53) introduces the ability to view container logs for the running instance. We need to add a UI component to allow the user to view the logs when the instance is present. This could be an option on the Tale menu unless there's a better place to put it.

bodom0015:
## Problem
There is no way to view the logs for a running Tale Instance from the WholeTale Dashboard.

Fixes #308

## Approach
* Add a call to `/instance/{id}/log` within the `InstanceService`
* Abstract `view-logs-dialog` to work not only with a list of Job ids, but also with an instance id
    * if given an `Instance` resource as `this.data.instance`, the dialog will show the logs for that Instance
    * otherwise, it will assume the previous behavior and will attempt to read from `this.data.jobIds` to fetch/merge those logs accordingly

Screen Shot 2022-12-05 at 2 17 50 PM

Screen Shot 2022-12-05 at 2 15 41 PM

## How to Test
Prerequisites: pull from upstream for all source repos, at least one Tale created/launched

3. Navigate to the Tale Dashboard
4. Choose a Tale that your own, hover over it, and click View (or click on the Tale title)
    * You should be brought to the Run Tale page
5. Start the Tale, if you haven't already done so
6. Expand the Tale dropdown menu at the top-right
    * You should see a new option here to "View Logs"
7. Click the "View Logs" button
    * You should see the familiar "View Logs" dialog, but now reading the logs for the current Instance of the chosen Tale


craig-willis:
This is looking good.  Can we hide the menu option when there isn't an instance? Currently the menu is there and does nothing.

bodom0015:
@craig-willis absolutely :+1: good call!

I pushed a fix to hide "View Logs" if there's no instance or status indicates that it is not running

craig-willis:
That works. 

While testing your change I came across another minor issues.  It looks like RStudio doesn't produce any logs during startup.  When hitting `instance/{id}/log` for a running RStudio tale, the response body has no content. The View Logs window shows "Fetching logs, please wait...".

bodom0015:
@craig-willis this is a known issue with the build logs as well... we currently assume that an empty array of `logs` messages means that we haven't fetched them yet.

If this is a case that we should support properly in the UI, I can try to put a "No logs found" message or similar there when it gets back a response that is empty?

craig-willis:
That sounds reasonable to me.

craig-willis (approved):

Xarthisius (approved):

PR #325, opened 2023-04-21
feat: use collections for ext data if catalog is enabled

Xarthisius:
## Problem

Currently "add data" modal only works with registered data. However, some WT deployments use data managed directly by girder. There's no easy way to add that data to Tales in that case.

## Approach
Allows to use native Girder data in Tales if `wholetale.enable_data_catalog` is set to True

## How to Test

3. Create a Tale
4. Register some external data, verify that 'add data' modal works.
5. Navigate to Girder. In Wholetale plugin settings, enable Data Catalog. Create some collections and populate them with data.
6. Go back to dashboard, verify that 'add data' modal now works on collections


craig-willis:
Looks good.  Just tracking for posterity that if `wholetale.enable_data_catalog` is true:
* Register menu is disabled
* My External Data is disabled

As discussed, for now we can't have a combination of registered data and catalog data. (I also see the complexity of browsing data from the catalog via a list of collections v registered data as a list of folders in a collection. Also that it isn't possible to add a collection -- e.g., JHU Seedling).

Xarthisius:
> Also that it isn't possible to add a collection -- e.g., JHU Seedling).

Yeah, that part I had to explicitly enforce on the dashboard side. DMS won't handle a Collection atm. That'd be a bigger change.

PR #333, opened 2024-06-20
Add a sunset banner

Xarthisius:
## Problem

We need to notify users that https://dashboard.wholetale.org is going offline and let them know how to get their data out.

## Approach

Show a simple banner whenever user logs in.

## How to Test

1. Set `BANNER=true` on dashboard container in docker-stack.yml
2. Checkout this branch locally, rebuild the dashboard
3. Login to the WholeTale Dashboard.
4. Banner should appear:

sunset

I'm open to suggestions regarding the message.


craig-willis:
Good idea. I'll likely revise the text and also point to a separate page on the docs site describing the shutdown.

craig-willis:
A drafted something in the docs. Not much different, but we can modify without changing the dashboard code https://github.com/whole-tale/wt-design-docs/pull/181.

